PR #16, opened 2015-10-30
Fix vpImageTools::binarise(), add testImageBinarise.cpp and testPerformanceLUT.cpp.

s-trinh:
Fix vpImageTools::binarise() function to be able to work with other types than unsigned char. 
Add test file for vpImageTools::binarise() and performance file for LUT vs iterate methods.
Add default warning function when trying to call vpImage<>.performLut() on image data types different than unsigned char or vpRGBa.


s-trinh:
Close this PR to avoid to commit multiple merges.

PR #48, opened 2016-02-18
Fix vpPoseVector::getThetaUVector() to return the adequate vector

jchevrie:
vpPoseVector::getThetaUVector() was returning a vpThetaUVector with the values corresponding to the translation instead of the angle/axis representation.


fspindle:
Merci

PR #56, opened 2016-03-14
Fix build errors when flag BUILD_DEPRECATED_FUNCTIONS is set to OFF.

jchevrie:
Fix build errors when flag BUILD_DEPRECATED_FUNCTIONS is set to OFF.

Mainly add missing preprocessor "#if defined(VISP_BUILD_DEPRECATED_FUNCTIONS)"


fspindle:
Thanks

PR #85, opened 2016-09-20
Fix for morphology operations + add morphology in the general case.

### LINKED ISSUE #84: Problem with vpImageMorphology and CONNEXITY_8 and with border pixels.
In the current code, the column indexes are wrong for [erosion](https://github.com/lagadic/visp/blob/3.0.0/modules/core/include/visp3/core/vpImageMorphology.h#L144-L146) and for [dilatation](https://github.com/lagadic/visp/blob/3.0.0/modules/core/include/visp3/core/vpImageMorphology.h#L208-L210).

The pixels located in the border are not taken into account also.

s-trinh:
Fix morphology operations on binary images (resolve #84 ). 
Add morphology operations in the general case (grayscale images) with naive SSE implementation.

The test file is long as we have to check the results for:

- erosion 4-connexity on binary images

- erosion 8-connexity on binary images

- dilatation 4-connexity on binary images

- dilatation 8-connexity on binary images

- same things for grayscale images

- plus performance comparison between regular and naive SSE code

Matlab R2014b is used to get the reference results with [`imdilate`](https://www.mathworks.com/help/images/ref/imdilate.html) and [`imerode`](https://www.mathworks.com/help/images/ref/imerode.html) functions.

### DIFF
--- a/modules/core/test/image/testImageMorphology.cpp
+++ b/modules/core/test/image/testImageMorphology.cpp
@@ -1,0 +1,847 @@
+/****************************************************************************
+ *
+ * This file is part of the ViSP software.
+ * Copyright (C) 2005 - 2015 by Inria. All rights reserved.
+ *
+ * This software is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * ("GPL") version 2 as published by the Free Software Foundation.
+ * See the file LICENSE.txt at the root directory of this source
+ * distribution for additional information about the GNU GPL.
+ *
+ * For using ViSP with software that can not be combined with the GNU
+ * GPL, please contact Inria about acquiring a ViSP Professional
+ * Edition License.
+ *
+ * See http://visp.inria.fr for more information.
+ *
+ * This software was developed at:
+ * Inria Rennes - Bretagne Atlantique
+ * Campus Universitaire de Beaulieu
+ * 35042 Rennes Cedex
+ * France
+ *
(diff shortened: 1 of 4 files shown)


modules/core/test/image/testImageMorphology.cpp
```diff
@@ -0,0 +1,847 @@
...
+          max_value = std::max(max_value, J.bitmap[i*J.getWidth() + j + offset[k]]);
+        }
+
+        I.bitmap[i*I.getWidth() + j] = max_value;
+      }
+    }
+  }
+}
+
+int modulo(const int a, const int n) {
```
fspindle: I suggest to add modulo in vpMath

modules/core/test/image/testImageMorphology.cpp
```diff
@@ -0,0 +1,847 @@
...
+      138,138,157,176,195,214,233,252,255,255,59,59,59,78,79,98,117,
+      156,156,175,194,213,232,251,255,255,255,59,59,59,78,97,116,135
+    };
+
+    vpImage<unsigned char> I2_check_eroded2(image_data2_eroded2, 17, 17, true);
+
+    if (I2_check_eroded2 != I_magic_square_erosion2_sse) {
+      throw vpException(vpException::fatalError, "(I2_check_eroded2 != I_magic_square_erosion2_sse)");
+    }
+
```
fspindle: Please keep only one empty line. Seen also before and after.

PR #93, opened 2016-10-25
Update vpRealSense to be able to handle multiple stream format.

s-trinh:
Work in progress.

It is possible to configure the streams directly using the `rs::device` obtained by `getHandler()`.
So the proposed changes are just a more convenient way to configure the streams.
The proposed changes can be discussed / reverted.

### DIFF
--- a/modules/sensor/include/visp3/sensor/vpRealSense.h
+++ b/modules/sensor/include/visp3/sensor/vpRealSense.h
@@ -39,6 +39,7 @@
 #define __vpRealSense_h_
 
 #include <stdint.h>
+#include <map>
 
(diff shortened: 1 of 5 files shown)


fspindle:
Ok for the principle, but changes that are proposed are too much SR300 centered. We should find a way to be more generic.

s-trinh:
- [x] remove SR300 presets

- [x] remove added `acquire()` 

- [x] replaced by generic `acquire()` which allows to copy directly from the native stream buffer. The developer has to take care to the proper type/dimension of the array.

- [ ] propagate changes to PCL sections

- [ ] check with other data type

- [ ] add example code + documentation

PR #100, opened 2016-11-17
Add extended vpThread example.

s-trinh:
Show a way to use multiple arguments and return values with vpThread.


s-trinh:
Close, need to check something.

PR #137, opened 2017-03-03
Fix bug in extration of vpRotationMatrix from vpPoseVector using vpPo…

jchevrie:
…seVector::getRotationMatrix()


fspindle:
Argg

Thanks

PR #151, opened 2017-03-19
First and last image determination

yurriy:
vpVideoReader :: findFirstFrameIndex(), ::findLastFrameIndex() functions now read directory files, where images must be located, and extract the lowest/highest index of image with specified format.

### DIFF
--- a/modules/core/include/visp3/core/vpIoTools.h
+++ b/modules/core/include/visp3/core/vpIoTools.h
@@ -195,6 +195,7 @@ public:
   static bool isSamePathname(const std::string& pathname1, const std::string& pathname2);
   static std::pair<std::string, std::string> splitDrive(const std::string& pathname);
   static std::vector<std::string> splitChain(const std::string & chain, const std::string & sep);
+  static std::vector<std::string> getDirFiles(const std::string &dirname);
 
   /*!
     @name Configuration file parsing
--- a/modules/core/src/tools/file/vpIoTools.cpp
+++ b/modules/core/src/tools/file/vpIoTools.cpp
@@ -1624,10 +1624,9 @@ std::vector<std::string> vpIoTools::splitChain(const std::string & chain, const 
  */
 std::vector<std::string> vpIoTools::getDirFiles(const std::string &pathname) {
 
-  std::string dirName = path(pathname);
-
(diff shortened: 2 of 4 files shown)


fspindle:
Nice. Thanks. Works well under OSX. I will check tomorrow under WIndows/Ubuntu.

My comment concerns sscanf() usage:
- Could you please remove sscanf() usage that brings warning with static analysis tools like cppcheck or Coverity Scan. 
- It should be possible to do the same in c++.

PS: Following this work, I think that vpVideoReader could be improved more (see #154 and #155).

fspindle:
Sorry but image format like `I%d.pgm` or `I%10d.pgm` (with digit greater than 9) are not supported. 

    vpException: Format of image name wasn't recognized: I%10d.pgm

yurriy:
@fspindle, thanks, fixed.
Note that the behavior is a little bit more restricted, than it was with sscanf: now it checks that image's index has the length exactly which was specified between % and d, or that it doesn't contain leading zeros, if specifier "%d" was used.

fspindle:
Sounds ok for me.

Able to consider %d and %0xd names

@Souriya Any remark ?

s-trinh:
Looks also ok for me.

@yurriy Maybe you could add a note to clarify the behavior of `getDirFiles()`?

- looks like it returns the list of files and subdirectories (tested on Ubuntu, it includes also `..` and `.`?)

- with no predefined order (tested on Ubuntu)

fspindle:
Thanks Souriya you are right.

@yurriy  I propose that you remove `..` and `.` from the list returned by `getDirFiles()`

fspindle:
Doesn't build under windows with msvc 2017.
There is `cctype` include that defines `std::isdigit()` that is missing.
Just add `#include <cctype>` in `vpVideoReader.cpp`

fspindle:
Works under Windows but you should also remove `.` and `..`
Once done I will merge.

yurriy:
@fspindle 
Does it still return . and ..? I thought I removed it in bf6c5b52de39bd6cde063ba6e67482e1fed44cc9

fspindle:
Yes under windows it returns also `.`and `..`
What you have introduced in bf6c5b5 is only for unix. The same should be done for windows.

fspindle:
Thanks

PR #164, opened 2017-03-29
Simplify some code. Add small optimizations.

s-trinh:
- Simplify some code in `vpColVector` and `vpMatrix` for insertion.

- Avoid sorting all the vectors when the size if even in `vpMath::getMedian()`

### DIFF
--- a/modules/core/src/math/matrix/vpMatrix.cpp
+++ b/modules/core/src/math/matrix/vpMatrix.cpp
@@ -4470,12 +4470,10 @@ void vpMatrix::insert(const vpMatrix&A, const unsigned int r,
   if( (r + A.getRows() ) <= rowNum && (c + A.getCols() ) <= colNum ){
     if (A.colNum == colNum && data != NULL && A.data != NULL && A.size() > 0) {
       memcpy(data+r*colNum, A.data, sizeof(double)*A.size());
-    } else {
+    } else if(data != NULL && A.data != NULL && A.colNum > 0) {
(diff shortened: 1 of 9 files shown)


modules/core/src/math/matrix/vpMatrix.cpp
```diff
@@ -2921,10 +2920,14 @@ void vpMatrix::insert(const vpMatrix&A, const unsigned int r,
...
-    for(unsigned int i=r; i<(r+A.getRows()); i++){
-      for(unsigned int j=c; j<(c+A.getCols()); j++){
-        (*this)[i][j] = A[i-r][j-c];
+    if (A.colNum == colNum && data != NULL && A.data != NULL && A.size() > 0) {
+      memcpy(data+r*colNum, A.data, sizeof(double)*A.size());
+    } else {
+      // recopy matrix A in the current one, does not call static function to avoid initialisation and recopy of matrix
+      for(unsigned int i=r; i<(r+A.getRows()); i++){
+        for(unsigned int j=c; j<(c+A.getCols()); j++){
+          (*this)[i][j] = A[i-r][j-c];
```
fspindle: Maybe here we could do a `memcpy()` for each line

s-trinh: Yes indeed. It should be possible to improve `vpMatrix::stack()` also. I will continue to push the developments in this branch.

fspindle:
Nice

Issue #154, opened 2017-03-20
Improve vpVideoReader to be able to get first and last frame without calling open()


fspindle:
In PR #151 getFirstFrame() and getLastFrame were improved, but these methods can only be used after a call to open() like in the following code:

````
#include <visp3/visp.h>

int main(int argc, const char *argv[])
{
  vpImage<unsigned char> I;
  vpVideoReader grabber;
  grabber.setFileName(argv[1]);
  grabber.open(I);
  std::cout << "First: " << grabber.getFirstFrameIndex() << " last: " << grabber.getLastFrameIndex() << std::endl;
}
````

A nice improvement would be to be able to get first and last frame number of a sequence of images without calling open() like:

````
#include <visp3/visp.h>

int main(int argc, const char *argv[])
{
  vpVideoReader grabber;
  grabber.setFileName(argv[1]);
  std::cout << "First: " << grabber.getFirstFrameIndex() << " last: " << grabber.getLastFrameIndex() << std::endl;
}
````

yurriy:
I'll do it

yurriy:
It must work only for image sequences?

fspindle:
Good question. 
In this enhancement we can consider only image sequences.

PR #186, opened 2017-06-10
frames indices calculation without open() call

yurriy:
Improvement from issue #154

### DIFF
--- a/modules/io/include/visp3/io/vpVideoReader.h
+++ b/modules/io/include/visp3/io/vpVideoReader.h
@@ -283,18 +283,9 @@ public:
     */
     inline long getFrameIndex() const { return frameCount;}
 
-    /*!
-      Gets the first frame index.
+    long getFirstFrameIndex();
+    long getLastFrameIndex();
 
-      \return Returns the first frame index.
-    */
-    inline long getFirstFrameIndex() const {return firstFrame;}
-    /*!
-      Gets the last frame index.
-
-      \return Returns the last frame index.
-    */
-    inline long getLastFrameIndex() const {return lastFrame;}
     /*!
       Gets the frame step.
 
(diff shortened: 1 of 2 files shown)


fspindle:
Works well for an image sequence, but not with a video file (ie .mp4 handled with OpenCV). I used the following code for testing:

````
#include <iostream>
#include <visp3/visp.h>

int main(int argc, const char *argv[])
{
  try {
    if (argc == 1) {
      std::cout << "Usage: " << argv[0] << " <image sequence>" << std::endl;
      return 0;
    }
    vpImage<unsigned char> I;
    vpVideoReader grabber;
    grabber.setFileName(argv[1]);
    if (argc > 2) {
      std::cout << "Call open()" << std::endl;
      grabber.open(I);
    }
    std::cout << "First: " << grabber.getFirstFrameIndex() << " last: " << grabber.getLastFrameIndex() << std::endl;
  }
  catch(const vpException &e) {
    std::cout << "Catched exception: " << e.getStringMessage() << std::endl;
  }
}
````

and here the result:

````
$ ./get-first-last-frame ../video/video-1-24.mp4 
First: 0 last: 0
````

yurriy:
It must return -1 for video files?

fspindle:
Not -1 but 1L for first frame and the value of the last frame.

What I suggest:
- create a new private function `void getProperties()` that contains more or less what is in `open()` and that updates the fisrt, last frame number, the image size and set isOpen to true
- Then `open()` becomes 
````
void vpVideoReader::open(vpImage<unsigned char> &I) {
  getProperties();
  I.resize(height, width);
}
````
- modify `getFirstFrame()` with something similar to:
````
getFirstFrame() {
  if (!isOpen) {
    getProperties();
  }
  return firstFrame;
}
````
and similarly for `getLastFrame()` with something similar to:
````
getLastFrame() {
  if (!isOpen) {
    getProperties();
  }
  return lastFrame;
}
````

fspindle:
Based on my previous comment I have introduced a more generic way to get video properties (first and last frame, framerate) in PR #215

PR #199, opened 2017-07-04
Introduce dgemm and dgemv usage with lapack built-in

fspindle:


fspindle:
Doesn't allow to optimize the code since it uses functions converted from fortran to C.

PR #210, opened 2017-07-24
AprilTag first integration into ViSP

s-trinh:
Proposal for AprilTag integration into ViSP.
Review to be done.
Unix only currently as:
- `pthread` is used in AprilTag
- `#include <sys/time.h>`
- some other errors when compiling with Visual Studio, Mingw

Result (pose from homography, `quad_decimate=2`):
image

Result (pose from homography, `quad_decimate=1`):
image

Note: 
- the detection can be tuned (reduce the computation time, increase the detection, ...)
- two methods for pose: ViSP `PnP` method or pose from homography
- add the pose from homography to initialize the ViSP VVS pose estimation method?
- the computation time displayed corresponds to a one shot detection, the computation time should be better once the detector is initialized


fspindle:
Nice work. Ok on OSX.

- Concerning the pose computation, using ViSP PnP should lead to a better precision since distorsion could be taken into account. Adding the pose from homography to initialize the ViSP VVS pose estimation method is a good idea. Since pose computation is very fast, we could do pose estimation from homography + Lagrange + Dementhon and then use the one with the smallest residual to initialize VVS. Finally, what I see, is only one method for pose estimation:
````
if (computePose) {
  // - homography
  // - Lagrange
  // - Dementhon
  // - check residual
  // - VVS
}
````
- Concerning the port to windows, we should check if <https://www.sourceware.org/pthreads-win32/> that is LGPL (compatible with GPL) could be introduced as a built-in 3rd party

**Comment by fspindle** [comment]
@s-trinh Wonderful. Very good work.

I have some suggestions:
- we should add a `3rdparty/apriltag/README.ViSP` that should mention the version that was introduced. In this file we should also indicate the changes that were done to port apriltag to windows. This could be done giving the link to the corresponding commits
- in `3rdparty/apriltag` we should also add a `LICENSE`file with the content of BSD clause-2
- to have the same philosophy than existing 3rd parties, we should remove the intermediate folder  `pthreads-w32-2-9-1-release` to have  `3rdparty/pthreads-win32`. Adding here also README.VISP that indicates that it is version w32-2-9-1-release could be nice.

**Comment by s-trinh** [comment]
Currently, I have added [here](https://github.com/s-trinh/visp/blob/7c533f922df7624cb0a578cb8bf75af8b5f6bff8/3rdparty/apriltag/CMakeLists.txt#L10-L13):
```
add_library(${APRILTAG_LIBRARY} STATIC ${lib_srcs} ${lib_hdrs})
if(WIN32 AND MSVC)
  target_link_libraries (${APRILTAG_LIBRARY} ${PTHREADS_LIBRARY})
endif()
```
as AprilTag needs pthreads. I don't think it is the right way to do.
Also, `visp_pthreads.dll` is created in the `lib` directory not in the `bin` directory with this method and is not installed.
The CMake part will need some modifications.

s-trinh:
We have [travis](https://travis-ci.org/) for continuous integration for Unix. Maybe we should look also at [AppVeyor](https://www.appveyor.com/) for Windows builds?
[Here](https://github.com/google/protobuf/blob/master/appveyor.yml) an example with Google Protobuf.

fspindle:
Thanks Souriya.

I will check the Windows port after my holidays.
Yes I agree, we should also use something like AppVeyor for windows.

PR #211, opened 2017-07-26
Add Basler cameras support with Pylon SDK

xoox:
Similar to vpFlyCaptureGrabber, a new abstract class vpPylonGrabber and
subclasses are added, which are much simplified clones of
vpFlyCaptureGrabber.

To test this PR, install first Pylon SDK.
https://www.baslerweb.com/en/support/downloads/software-downloads/#type=pylonsoftware;version=all

Windows OSes are not supported yet by FindPylon.cmake module.
So far this PR was tested under Ubuntu with pylon 5.0.9 and the
following GigE cameras:
- acA640-90gm
- acA1600-60gm


fspindle:
Nice contrib. Thanks.
Build tested on Ubuntu 17.04 and OSX.

@s-trinh I think that we can merge this PR. Support for Windows could be added later, after my vacation

PR #220, opened 2017-08-31
Fix AprilTag struct initialization.

s-trinh:


fspindle:
Requested to have a successful build on Ubuntu 12.04 i386

PR #241, opened 2017-10-09
sorting result of vpIoTools::getDirFiles()

MarcPouliquenInria:
Adding a sort of the vector containing all the files in a directory,  to have a vector sorted in alphabetical order in output


modules/core/src/tools/file/vpIoTools.cpp
```diff
@@ -1705,6 +1705,7 @@ std::vector<std::string> vpIoTools::getDirFiles(const std::string &pathname) {
     free(list[i]);
   }
   free(list);
+  std::sort(files.begin(), files.end());
   return files;
```
fspindle: Thanks, but could you also add the `std::sort` line for Windows case (near line 1734)

MarcPouliquenInria: Yes, it's done in next commit

PR #269, opened 2017-12-06
Fix bug: serial printed as string

xoox:
The serial number in pylon library is stored as a string. When printing
out, it has to be formatted as string characters instead of unsigned
decimal integer.


fspindle:
Thanks

PR #288, opened 2018-01-16
Fix for Win XP inet_ntop() could not be located in WS2_32.dll

### LINKED ISSUE #287: On Win XP inet_ntop() could not be located in WS2_32.dll
There is an error during execution of any test that needs `visp_core`:

`The procedure entry point inet_ntop() could not be located in the dynamic link library WS2_32.dll`.

This error occurs on Win XP with Visual 2010

fspindle:
Make vpNetwork, vpClient, vpServer classes available only since win 8.1 where inet_ntop() is implemented. Closes #287

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ ViSP 3.2.0 (under development)
   - Bug fixed
     . [#279] Unable to build ViSP with Visual Studio 2010
     . [#281] Segfault in mbtGenericTrackingDepth.cpp when run on big endian arch
+    . [#287] On Win XP inet_ntop() could not be located in WS2_32.dll
 
 ----------------------------------------------
 ViSP 3.1.0 (released December 22nd, 2017)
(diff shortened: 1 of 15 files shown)


s-trinh:
Small precision: I am able to run these tests on Windows 7 with Visual Studio 2015.
It probably depends on the version of Windows SDK (I have Windows Kits 8.0, 8.1, 10) rather than the OS.
For instance, `inet_ntop` is defined here for VS2015 on my computer: `c:\Program Files (x86)\Windows Kits\8.1\Include\um\WS2tcpip.h`.

fspindle:
Argg. My changes are then wrong.
From the [official doc](https://msdn.microsoft.com/fr-fr/library/windows/desktop/ms738564(v=vs.85).aspx) I understood that it is the OS version, nor the SDK. [Here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx) are the values of `_WIN32_WINRT` corresponding to the OS.

Tomorrow I will check again the value of `_WIN32_WINRT` macro on my win XP

PR #289, opened 2018-01-18
update doc

montanier:
The original documentation proposes to set the initial position of the object to 0,0,1. This is likely to fail at initialization time when the method "outOfImage" of the class "vpMeTracker" is called to check if the points of the object are within the zone of attention (the image minus some margin).

Proposing an initialization to another value (less close to 0, and therefore within the zone of attention of the tracker) would avoid to run in this error.


fspindle:
Thanks for trying to contribute, but the proposed PR could not be accepted.
- The initial position set to `0,0,1,0,0,0` was just an example and should be considered as is
- The values allow to center the object in the image if the origin of the object frame is located in the middle of the object
- Using  `0.1,0.1,1,0,0,0` brings a translation shift that could center your object in the image only if the object frame is not located in the middle of the object
- To conclude, values that have to be used cannot be generic, but adapted to each CAD model depending especially on the location (position and orientation) of the object frame

Issue #304, opened 2018-03-06
Query regarding GSOC project


AKS1996:
Hi

I'm Akshay. I've been trying to get your attention for the GSOC project **Porting ViSP to Android**. I contacted your org on the google group a week ago but it must have missed your attention. It'll be nice if I can get in touch with someone from your org since the proposal date is coming soon.

In this [link](https://groups.google.com/forum/m/#!topic/visp-gsoc-2018/jMk_iIDIBrA), I've mentioned my past experience and had some queries regarding the idea.

Thanks  @fspindle @MarcPouliquenInria

PR #307, opened 2018-03-12
Porting ViSP to android

AKS1996:
Hi @fspindle 

As per as previous conversation #304, I've learnt how to build a C++ package for android platform using ```android.toolchain.cmake``` and I'm on my way to build OpenCV for Android(having some 3rd party library issues, you can get the source code [here](https://github.com/AKS1996/GSOC-Prep)). 

Meanwhile I began adding android build flags to ViSP's cmake files. As you said, I'm referring OpenCV's cmake files.

There are few ViSP libraries for which there's no corresponding cmake file in the OpenCV source, which I could've referred to. So it'll take time to find such libraries and edit their cmake files.
Also I wanted to have quick android demo showing any of the ViSP's module. But I don't know which one to start with, which is independent of other modules - so as make least edits.

I'll appreciate any advice from you regarding the edits made and whether I'm going in right direction or not.

Thanks
Akshay

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,10 +42,12 @@ if(NOT CMAKE_TOOLCHAIN_FILE)
   if(WIN32)
     set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "Installation Directory")
   endif()
-else()
+else(NOT CMAKE_TOOLCHAIN_FILE)
+  #Android: set output folder to ${CMAKE_BINARY_DIR}
+  set( LIBRARY_OUTPUT_PATH_ROOT ${CMAKE_BINARY_DIR} CACHE PATH "root for library output, set this to change where android libs are compiled to" )
   # Crosscompiling
   set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "Installation Directory")
-endif()
+endif(NOT CMAKE_TOOLCHAIN_FILE)
 
 if(CMAKE_SYSTEM_NAME MATCHES WindowsPhone OR CMAKE_SYSTEM_NAME MATCHES WindowsStore)
   set(WINRT TRUE)
--- a/cmake/AddExtraCompilationFlags.cmake
+++ b/cmake/AddExtraCompilationFlags.cmake
@@ -136,7 +136,7 @@ endif()
 
 if(UNIX)
   if(CMAKE_COMPILER_IS_GNUCXX)
-    add_extra_compiler_option(-fPIC)
+    add_extra_compiler_option(-fPIC) # Is needed for ANDROID too.
   endif()
 endif()
 
--- a/cmake/DetectAndroidSDK.cmake
+++ b/cmake/DetectAndroidSDK.cmake
@@ -208,13 +208,12 @@ macro(add_android_project target path)
                      The project ${target} will not be build")
   endif()
 
-# TODO to be edited    
-#  # check native dependencies
-#  if(android_proj_IGNORE_JAVA)
-#    ocv_check_dependencies(${android_proj_NATIVE_DEPS})
-#  else()
-#    ocv_check_dependencies(${android_proj_NATIVE_DEPS} opencv_java)
-#  endif()
+  # check native dependencies
+  if(android_proj_IGNORE_JAVA)
+    vp_check_dependencies(${android_proj_NATIVE_DEPS})
+  else()
+    vp_check_dependencies(${android_proj_NATIVE_DEPS} opencv_java)
+  endif()
 
   if(EXISTS "${path}/jni/Android.mk" )
     # find if native_app_glue is used
@@ -348,7 +347,7 @@ macro(add_android_project target path)
     endif()
     set(__android_project_chain ${target} CACHE INTERNAL "auxiliary variable used for Android progects chaining")
 
-    # put the final .apk to the OpenCV's bin folder
+    # put the final .apk to the VISP's bin folder
     add_custom_command(TARGET ${target} POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different "${android_proj_bin_dir}/bin/${target}-debug.apk" "${VISP_BINARY_DIR}/bin/${target}.apk")
     if(INSTALL_ANDROID_EXAMPLES AND "${target}" MATCHES "^example-")
       #apk
--- a/cmake/DetectAndroidSDK.cmake
+++ b/cmake/DetectAndroidSDK.cmake
@@ -1,0 +1,381 @@
+if(EXISTS "${ANDROID_EXECUTABLE}")
+  set(ANDROID_SDK_DETECT_QUIET TRUE)
+endif()
+
+file(TO_CMAKE_PATH "$ENV{ProgramFiles}" ProgramFiles_ENV_PATH)
+file(TO_CMAKE_PATH "$ENV{HOME}" HOME_ENV_PATH)
+
+if(CMAKE_HOST_WIN32)
+  set(ANDROID_SDK_OS windows)
+elseif(CMAKE_HOST_APPLE)
+  set(ANDROID_SDK_OS macosx)
+else()
+  set(ANDROID_SDK_OS linux)
+endif()
+
+#find android SDK: search in ANDROID_SDK first
+find_host_program(ANDROID_EXECUTABLE
+  NAMES android.bat android
+  PATH_SUFFIXES tools
+  PATHS
+    ENV ANDROID_SDK
+  DOC "Android SDK location"
+  NO_DEFAULT_PATH
+  )
+
+# Now search default paths
+find_host_program(ANDROID_EXECUTABLE
+  NAMES android.bat android
+  PATH_SUFFIXES android-sdk-${ANDROID_SDK_OS}/tools
+                android-sdk-${ANDROID_SDK_OS}_x86/tools
+                android-sdk-${ANDROID_SDK_OS}_86/tools
+                android-sdk/tools
+  PATHS /opt
+        "${HOME_ENV_PATH}/NVPACK"
+        "$ENV{SystemDrive}/NVPACK"
+        "${ProgramFiles_ENV_PATH}/Android"
+  DOC "Android SDK location"
+  )
+
+if(ANDROID_EXECUTABLE)
+  if(NOT ANDROID_SDK_DETECT_QUIET)
+    message(STATUS "Found android tool: ${ANDROID_EXECUTABLE}")
+  endif()
+
+  get_filename_component(ANDROID_SDK_TOOLS_PATH "${ANDROID_EXECUTABLE}" PATH)
+
+  #read source.properties
+  if(EXISTS "${ANDROID_SDK_TOOLS_PATH}/source.properties")
+    file(STRINGS "${ANDROID_SDK_TOOLS_PATH}/source.properties" ANDROID_SDK_TOOLS_SOURCE_PROPERTIES_LINES REGEX "^[ ]*[^#].*$")
+    foreach(line ${ANDROID_SDK_TOOLS_SOURCE_PROPERTIES_LINES})
+      string(REPLACE "\\:" ":" line ${line})
+      string(REPLACE "=" ";" line ${line})
+      list(GET line 0 line_name)
+      list(GET line 1 line_value)
+      string(REPLACE "." "_" line_name ${line_name})
+      SET(ANDROID_TOOLS_${line_name} "${line_value}" CACHE INTERNAL "from ${ANDROID_SDK_TOOLS_PATH}/source.properties")
+      MARK_AS_ADVANCED(ANDROID_TOOLS_${line_name})
+    endforeach()
+  endif()
+
+  #fix missing revision (SDK tools before r9 don't set revision number correctly)
+  if(NOT ANDROID_TOOLS_Pkg_Revision)
+    SET(ANDROID_TOOLS_Pkg_Revision "Unknown" CACHE INTERNAL "")
+    MARK_AS_ADVANCED(ANDROID_TOOLS_Pkg_Revision)
+  endif()
+
+  #fix missing description
+  if(NOT ANDROID_TOOLS_Pkg_Desc)
+    SET(ANDROID_TOOLS_Pkg_Desc "Android SDK Tools, revision ${ANDROID_TOOLS_Pkg_Revision}." CACHE INTERNAL "")
+    MARK_AS_ADVANCED(ANDROID_TOOLS_Pkg_Desc)
+  endif()
+
+  #warn about outdated SDK
+  if(NOT ANDROID_TOOLS_Pkg_Revision GREATER 13)
+    SET(ANDROID_TOOLS_Pkg_Desc "${ANDROID_TOOLS_Pkg_Desc} It is recommended to update your SDK tools to revision 14 or newer." CACHE INTERNAL "")
+  endif()
+
+  if(ANDROID_TOOLS_Pkg_Revision GREATER 13)
+    SET(ANDROID_PROJECT_PROPERTIES_FILE project.properties)
+    SET(ANDROID_ANT_PROPERTIES_FILE ant.properties)
+  else()
+    SET(ANDROID_PROJECT_PROPERTIES_FILE default.properties)
+    SET(ANDROID_ANT_PROPERTIES_FILE build.properties)
+  endif()
+
+  set(ANDROID_MANIFEST_FILE AndroidManifest.xml)
+  set(ANDROID_LIB_PROJECT_FILES build.xml local.properties proguard-project.txt ${ANDROID_PROJECT_PROPERTIES_FILE})
+  set(ANDROID_PROJECT_FILES ${ANDROID_LIB_PROJECT_FILES})
+
+  #get installed targets
+  if(ANDROID_TOOLS_Pkg_Revision GREATER 11)
+    execute_process(COMMAND ${ANDROID_EXECUTABLE} list target -c
+      RESULT_VARIABLE ANDROID_PROCESS
+      OUTPUT_VARIABLE ANDROID_SDK_TARGETS
+      ERROR_VARIABLE ANDROID_PROCESS_ERRORS
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+    string(REGEX MATCHALL "[^\n]+" ANDROID_SDK_TARGETS "${ANDROID_SDK_TARGETS}")
+  else()
+    #old SDKs (r11 and older) don't provide compact list
+    execute_process(COMMAND ${ANDROID_EXECUTABLE} list target
+      RESULT_VARIABLE ANDROID_PROCESS
+      OUTPUT_VARIABLE ANDROID_SDK_TARGETS_FULL
+      ERROR_VARIABLE ANDROID_PROCESS_ERRORS
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+    string(REGEX MATCHALL "(^|\n)id: [0-9]+ or \"([^\n]+[0-9+])\"(\n|$)" ANDROID_SDK_TARGETS_FULL "${ANDROID_SDK_TARGETS_FULL}")
+
+    SET(ANDROID_SDK_TARGETS "")
+    if(ANDROID_PROCESS EQUAL 0)
+      foreach(line ${ANDROID_SDK_TARGETS_FULL})
+        string(REGEX REPLACE "(^|\n)id: [0-9]+ or \"([^\n]+[0-9+])\"(\n|$)" "\\2" line "${line}")
+        list(APPEND ANDROID_SDK_TARGETS "${line}")
+      endforeach()
+    endif()
+  endif()
+
+  if(NOT ANDROID_PROCESS EQUAL 0)
+    message(ERROR "Failed to get list of installed Android targets.")
+    set(ANDROID_EXECUTABLE "ANDROID_EXECUTABLE-NOTFOUND")
+  endif()
+
+  # clear ANDROID_SDK_TARGET if no target is provided by user
+  if(NOT ANDROID_SDK_TARGET)
+    set(ANDROID_SDK_TARGET "" CACHE STRING "Android SDK target for the VISP Java API and samples")
+  endif()
+  if(ANDROID_SDK_TARGETS)
+    set_property( CACHE ANDROID_SDK_TARGET PROPERTY STRINGS ${ANDROID_SDK_TARGETS} )
+  endif()
+endif(ANDROID_EXECUTABLE)
+
+# finds minimal installed SDK target compatible with provided names or API levels
+# usage:
+#   get_compatible_android_api_level(VARIABLE [level1] [level2] ...)
+macro(android_get_compatible_target VAR)
+  set(${VAR} "${VAR}-NOTFOUND")
+  if(ANDROID_SDK_TARGETS)
+    list(GET ANDROID_SDK_TARGETS 0 __lvl)
+    string(REGEX MATCH "[0-9]+$" __lvl "${__lvl}")
+
+    #find minimal level mathing to all provided levels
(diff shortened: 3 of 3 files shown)


fspindle:
I think that we should start to port ViSP without the support of any 3rd party. There is already apriltag that is embedded in ViSP source code and that could be a good candidate for a first demo (see [tutorial](http://visp-doc.inria.fr/doxygen/visp-daily/tutorial-detection-apriltag.html)).

The pseudo-code of the demo could be:
````
while(1)
    android_image = acquire_image()
    convert_from_android_to_visp(android_image, visp_image)
    [nbtags, cog, roi] = detect_apriltag(visp_image)
    display_image(android_image)
    for (i=0; i < nbtags)
        display_roi(android_image, roi[i])
        display_cog(android_image, cog[i])
````

Than in a second time, we could add the support of OpenCV 3rd party.
And only later other 3rd parties like Eigen3 if possible.

**Comment by AKS1996** [comment]
hi @fspindle 

I'm on mid way adding and editing cmake files for android build. Since these files are much similar to the OpenCV analogues, I made a [README file](https://github.com/AKS1996/GSOC-Prep/tree/master/BuildVispAndroid/README.md) where you can refer the file changes I'm making.

Can you check:
+ whether I'm adding macros at the correct place or not
+ whether I'm trying to create a cmake script that already exists (with a different name perhaps) and provides the same code/functionality

thanks
akshay

**Comment by AKS1996** [comment]
hi @fspindle 

I had some doubts regarding the **core** module
+ It seems like the class `vpMatrix` needs at least one 3rd party library like Eigen, Lapack, OpenCV or GSL for building, [refer this line](https://github.com/lagadic/visp/blob/master/modules/core/src/math/matrix/vpMatrix.cpp#L1931)), else it throws an exception. So I'm using the Lapack library
+ Also the class `vpIOTools` requires `wordexp.h` header file, which is not available in android.

**Comment by fspindle** [comment]
Hi Akshay,

Thanks for the work you are doing. I will check your PR ASAP.

You are right, `vpMatrix` needs at least one 3d party. But if none is installed and detected we are using `clapack` that is [embedded](https://github.com/lagadic/visp/tree/master/3rdparty/clapack) in ViSP source code. This version is the result of fortran to c conversion. It is not optimized as the native one, but it does the trick. 

Concerning, `vpIoTools` I suggest to do the same as for iOS where `wordexp.h` is also not available.

````
#if !defined(_WIN32) && !defined(ANDROID)
#include <wordexp.h>
#endif
````
and later in the file something similar to
````
#if (TARGET_OS_IOS == 0) && !defined(ANDROID)
````

File: cmake/GenAndroidMK.cmake [hunk 1]
```diff
@@ -41,7 +41,7 @@ if (ANDROID)
                         set(lib_name "${output_name}")
                     endif ()
                 else ()
-                    ocv_get_libname(lib_name "${_output}")
+                    visp_get_libname(lib_name "${_output}")
```
fspindle: Should be here `vp_get_libname`

CMakeLists.txt
```diff
@@ -42,10 +42,12 @@ if(NOT CMAKE_TOOLCHAIN_FILE)
...
     set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "Installation Directory")
   endif()
-else()
+else(NOT CMAKE_TOOLCHAIN_FILE)
+  #Android: set output folder to ${CMAKE_BINARY_DIR}
+  set( LIBRARY_OUTPUT_PATH_ROOT ${CMAKE_BINARY_DIR} CACHE PATH "root for library output, set this to change where android libs are compiled to" )
   # Crosscompiling
   set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "Installation Directory")
-endif()
+endif(NOT CMAKE_TOOLCHAIN_FILE)
```
fspindle: Put `endif()`

CMakeLists.txt
```diff
@@ -42,10 +42,12 @@ if(NOT CMAKE_TOOLCHAIN_FILE)
   if(WIN32)
     set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "Installation Directory")
   endif()
-else()
+else(NOT CMAKE_TOOLCHAIN_FILE)
+  #Android: set output folder to ${CMAKE_BINARY_DIR}
+  set( LIBRARY_OUTPUT_PATH_ROOT ${CMAKE_BINARY_DIR} CACHE PATH "root for library output, set this to change where android libs are compiled to" )
   # Crosscompiling
```
fspindle: Remove space `set(LIBRARY_OUTPUT_PATH_ROOT ...)`

cmake/GenConfig.cmake
```diff
@@ -0,0 +1,132 @@
...
+if(INSTALL_TO_MANGLED_PATHS)
+  set(VISP_USE_MANGLED_PATHS_CONFIGCMAKE TRUE)
+else()
+  set(VISP_USE_MANGLED_PATHS_CONFIGCMAKE FALSE)
+endif()
+
+if(HAVE_CUDA)
+  vp_cmake_configure("${CMAKE_CURRENT_LIST_DIR}/templates/VISPConfig-CUDA.cmake.in" CUDA_CONFIGCMAKE @ONLY)
+endif()
+
```
fspindle: ViSP doesn't use CUDA.
Should be removed as well as `templates/VISPConfig-CUDA.cmake.in`

fspindle: I don't see the interest of this new file. There is already `cmake/VISPGenerateConfig.cmake`

cmake/GenConfig.cmake
```diff
@@ -0,0 +1,132 @@
...
+  if(ROOT_NAME)
+    # Root config file
+    configure_file("${VISP_SOURCE_DIR}/cmake/templates/${ROOT_NAME}" "${TMP_DIR}/VISPConfig.cmake" @ONLY)
+    install(FILES
+        "${TMP_DIR}/VISPConfig-version.cmake"
+        "${TMP_DIR}/VISPConfig.cmake"
+        DESTINATION "${VISP_CONFIG_INSTALL_PATH}" COMPONENT dev)
+  endif()
+endfunction()
+
```
fspindle: I see more this function in `cmake/VISPUtils.cmake`

cmake/VISPModule.cmake
```diff
@@ -775,8 +775,13 @@ macro(_vp_create_module)
     ARCHIVE_OUTPUT_DIRECTORY ${LIBRARY_OUTPUT_PATH}
     LIBRARY_OUTPUT_DIRECTORY ${LIBRARY_OUTPUT_PATH}
     RUNTIME_OUTPUT_DIRECTORY ${BINARY_OUTPUT_PATH}
+    DEFINE_SYMBOL CVAPI_EXPORTS
   )
```
fspindle: Is line 778 useful ? 
If yes it should be `VISP_EXPORTS`

cmake/VISPModule.cmake
```diff
@@ -775,8 +775,13 @@ macro(_vp_create_module)
...
 
+  if(ANDROID AND BUILD_FAT_JAVA_LIB)
+    target_compile_definitions(${the_module} PRIVATE CVAPI_EXPORTS)
+  endif()
```
fspindle: `VISP_EXPORTS` if needed

AKS1996:
hi @fspindle 

Thanks for the advice, made the edits in `vpIOTools`.

But soon after that, few files in `clapac` library were signalling undefined reference errors while linking. I [found the issue](https://github.com/nmoinvaz/minizip/issues/137) and edited a file named `fio.h` in that folder.

Now the build works well for `visp_core` module. You can get the install script [here](https://github.com/AKS1996/GSOC-Prep/tree/master/BuildVispAndroid)

Thanks 
akshay

cmake/VISPUtils.cmake
```diff
@@ -53,6 +53,52 @@ if(NOT COMMAND find_host_package)
...
+macro(vp_update VAR)
+  if(NOT DEFINED ${VAR})
+    if("x${ARGN}" STREQUAL "x")
+      set(${VAR} "")
+    else()
+      set(${VAR} ${ARGN})
+    endif()
+  else()
+    #ocv_debug_message("Preserve old value for ${VAR}: ${${VAR}}")
+  endif()
```
fspindle: Line 93 and 94 should be removed

cmake/VISPUtils.cmake
```diff
@@ -53,6 +53,52 @@ if(NOT COMMAND find_host_package)
...
+endmacro()
+
+# Finished adding macros for visp
+# -------------------------------------------
+
+
 # adds include directories in such way that directories from the ViSP source tree go first
```
fspindle: Remove empty line

platforms/android/build-tests/test_cmake_build.py
```diff
@@ -0,0 +1,135 @@
...
+int main(int argc, char* argv[])
+{
+  (void)argc; (void)argv;
+  printf("%s\\n", message);
+  Size textsize = getTextSize(message, CV_FONT_HERSHEY_COMPLEX, 3, 5, 0);
+  Mat img(textsize.height + 20, textsize.width + 20, CV_32FC1, Scalar(230,230,230));
+  putText(img, message, Point(10, img.rows - 10), CV_FONT_HERSHEY_COMPLEX, 3, Scalar(0, 0, 0), 5);
+  imwrite("/mnt/sdcard/HelloAndroid.png", img);
+  return 0;
+}
```
fspindle: Should be replaced with something that uses ViSP instead of OpenCV. Since we cannot write a text directly in a `vpImage`without using a `vpDisplay` we cannot do exactly the same as this example in ViSP. We could just create an image that contains a given gray level.

platforms/android/build-tests/test_cmake_build.py
```diff
@@ -0,0 +1,135 @@
...
+    def tearDown(self):
+        if os.path.exists(self.workdir):
+            shutil.rmtree(self.workdir)
+
+    def runTest(self):
+        cmd = [
+            "cmake",
+            "-GNinja",
+            "-DOpenCV_DIR=%s" % self.opencv_cmake_path,
+            "-DANDROID_ABI=%s" % self.abi,
```
fspindle: Should be `-DVISP_DIR`instead of `OpenCV_DIR`

platforms/android/build-tests/test_ndk_build.py
```diff
@@ -0,0 +1,148 @@
...
+    args = parser.parse_args()
+
+    if args.ndk_path is not None:
+        os.environ["ANDROID_NDK"] = os.path.abspath(args.ndk_path)
+
+    print("Using NDK: %s" % os.environ["ANDROID_NDK"])
+
+    res = unittest.TextTestRunner(verbosity=3).run(suite(os.path.abspath(args.workdir), os.path.abspath(args.opencv_mk_path)))
+    if not res.wasSuccessful():
+        sys.exit(res)
```
fspindle: Should be ported using ViSP

platforms/android/service/engine/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>OpenCVEngineService</name>
+	<comment></comment>
```
fspindle: `VISPEngineService`

platforms/android/service/engine/AndroidManifest.xml
```diff
@@ -0,0 +1,30 @@
...
+
+    <uses-sdk android:minSdkVersion="@ANDROID_NATIVE_API_LEVEL@" android:targetSdkVersion="22"/>
+    <uses-feature android:name="android.hardware.touchscreen" android:required="false"/>
+
+    <application
+        android:icon="@drawable/icon"
+        android:label="@string/app_name" android:allowBackup="true">
+
+    <service android:exported="true" android:name="OpenCVEngineService" android:process=":OpenCVEngineProcess">
+        <intent-filter>
```
fspindle: `VISP` and not `OpenCV`

platforms/android/service/engine/AndroidManifest.xml
```diff
@@ -0,0 +1,30 @@
...
+            <action android:name="org.opencv.engine.BIND"></action>
+        </intent-filter>
+    </service>
+
+    <activity
+        android:name="org.opencv.engine.manager.ManagerActivity"
+        android:label="@string/app_name"
```
fspindle: `visp`

platforms/android/service/engine/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/${ANDROID_MANIFEST_FILE}" "${OpenCV_BINARY_DIR}/platforms/android/service/engine/.build/${ANDROID_MANIFEST_FILE}"  @ONLY)
+unset(__android_project_chain CACHE)
+add_android_project(opencv_engine "${CMAKE_CURRENT_SOURCE_DIR}" SDK_TARGET 9 ${ANDROID_SDK_TARGET} IGNORE_JAVA ON IGNORE_MANIFEST ON COPY_LIBS ON)
```
fspindle: `visp`

platforms/android/service/engine/build.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project name="OpenCV Manager" default="help">
+
```
fspindle: `ViSP Manager`

platforms/android/service/engine/res/values/strings.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_name">OpenCV Manager</string>
+    <string name="about">About</string>
+    <string name="checkUpdate">Check for update</string>
+    <string name="intro">OpenCV library is used by other applications for image enhancement, panorama stitching, object detection, recognition and tracking and so on. OpenCV Manager provides the best version of the OpenCV for your hardware. See opencv.org for details.</string>
+</resources>
```
fspindle: `ViSP`

platforms/android/service/engine/src/org/opencv/engine/OpenCVEngineInterface.aidl
```diff
@@ -0,0 +1,33 @@
...
+    */
+    boolean installVersion(String version);
+
+    /**
+    * Return list of libraries in loading order separated by ";" symbol
+    * @param OpenCV version
+    * @return Returns OpenCV libraries names separated by symbol ";" in loading order
+    */
+    String getLibraryList(String version);
+}
```
fspindle: ViSP

platforms/android/service/engine/src/org/opencv/engine/manager/ManagerActivity.java
```diff
@@ -0,0 +1,113 @@
...
+                        + engine.getLibPathByVersion(null) + res);
+            } catch (RemoteException e) {
+                e.printStackTrace();
+                Log.e(TAG, "Call failed");
+            }
+            unbindService(this);
+        }
+    };
+
+}
```
fspindle: ViSP

platforms/android/service/readme.txt
```diff
@@ -0,0 +1,25 @@
...
+OpenCV installation on end user devices. In most cases OpenCV Manager may be installed automatically
+from Google Play.
+
+If Google Play is not available (i.e. on emulator, developer board, etc), you can install it
+manually using adb tool:
+
+    adb install <path-to-OpenCV-sdk>/apk/OpenCV_<version>_Manager_<app_version>_<platform>.apk
+
+Example: OpenCV_3.4.0-dev_Manager_3.40_armeabi-v7a.apk
+
```
fspindle: ViSP

CMakeLists.txt
AKS1996: done

cmake/GenConfig.cmake
AKS1996: transferred the function to VISPUtils.cmake

AKS1996:
Hi @fspindle 

Just finished making the OpenCV to VISP edits you mentioned. Also I'm trying to make a demo AprilTag detection app(get the source [code here](https://github.com/AKS1996/GSOC-Prep/tree/master/VispAndroidDemoApp)). I'll try to target only latest android version(API>=24) since older versions aren't maintained now and are buggy.

Following [the tutorial](http://visp-doc.inria.fr/doxygen/visp-daily/tutorial-detection-apriltag.html) you referred, creating a `vpImage` from android bitmap was easy. But I see that AprilTag detection needs a `vpImage<unsigned char>` while by default android provides `uint_16`(16 bit RGB-565 or 32 bit ARGB values). The conversion/copying is adding extra overheads while live tracking/detection(fails on high resolution images).

I read that OpenCV android sdk has its own [camera framework](https://github.com/steveliles/opencv-android/tree/master/OpenCV-Android/src/main/java/org/opencv/android)(they don't use the default one). I'll try to figure out how it works and update you.

Thanks
Akshay

fspindle:
Ok to consider the latest Android version.

In ViSP there is no aRGB to grey conversion, but if SSE3 is available on Android there is an optimized version of RGBa to grey [conversion](http://visp-doc.inria.fr/doxygen/visp-daily/classvpImageConvert.html#a29b1efabd1ae7f05927fea475cbb528e) that could be a a good starting to implement a new `vpImageConvert::aRGBToGrey()` function.

To speed up the Apriltag detection, before reconsidering a new camera framework, you can try to use:
- `vpDetectorAprilTag::setAprilTagQuadDecimate()` [function](http://visp-doc.inria.fr/doxygen/visp-daily/classvpDetectorAprilTag.html#a9fcd4eec9276f16a4d3a7689fb27696c): It operates like a subsampling factor. 
- `vpDetectorAprilTag::setAprilTagNbThreads()` [function](http://visp-doc.inria.fr/doxygen/visp-daily/classvpDetectorAprilTag.html#ae6e80854e8a882dea70ecc759eeb4ccc)

On 640 by 480 images, I succeeded to make detection realtime on Raspberry Pi 3 setting `quad_decimate=4` and `nthread=2`.

s-trinh:
Small precision, Android is most likely run on ARM platform and SSE intrinsics are x86 only. Maybe this conversion could be done directly with Android API?

AKS1996:
Hi all

Finally got a working demo for *single image AprilTag detection* on x86 arch, tested on API-24(get the [source code and prebuilt apk's here](https://github.com/AKS1996/GSOC-Prep/tree/master/VispAndroidDemoApp)). But there are some issues
+ The standard `bitmap.h` header in android NDK provides a `lockPixels` API, but it's [not reading pixel values correctly](https://github.com/AKS1996/GSOC-Prep/commit/da84c8040d40464831b4c63576dc6ab88af8d682)(I checked it against pixel values obtained on my PC). That's the reason I couldn't use `vpImageConvert::aRGBToGrey()` function. So I'm creating `vpImage` this way 
`bitmap(in java)` -> `byte array(in java)` -> `uchar array(in c++)` -> `vpImage (in c++)` -> 
+ While creating the image, I've keep the copy flag _true_ `vpImage(src, width, height, **true**)`, else it gives segmentation fault
+ On `armeabi-v7a` arch, it was giving some [undefined `atof` symbol](https://github.com/AKS1996/GSOC-Prep/tree/4688e6e35eafd18b7cf968ea002303b8f0a663d5) error. This is a problem either with the NDK or else we'll need to change the build procedure

I'm trying real time detection(`quad_decimate=4` proved to be good advice indeed), but it's still slow. I'll try to tweak a few things like FPS, or `vpDetector` attributes, etc and look for better results.

Lastly I had some doubts:
+ What should be the conversion `rgb2grey` function?
+ Now should I extend the app to multiple architectures(x86_64, mips) or focus on realtime detection on x86 only?

Looking for your advice
Akshay

AKS1996:
Hi all

Tweaked a few detection parameters(and some changes on Camera Preview on java side) and finally got a demo for realtime `vpAprilTag` detection with attributes:
+ x86 arch
+ API 25
+ 2 FPS, 640x480 greyscale images
+ `nbThreads=4` and `quad_decimate=4.0`

Get the demo [video here](https://www.youtube.com/embed/DIqEP6feL3I) and the [source code here](https://github.com/AKS1996/GSOC-Prep/tree/frame-grabbing/VispAndroidDemoApp)

Akshay

fspindle:
Thanks for the app demo. Very good work.

Regarding your previous message and questions, I would prefer that you try to improve the framerate of the demo instead of porting to multiple architectures that could be done later.  

I think that using `AndroidBitmap_lockPixels()` and `AndroidBitmap_unlockPixels()` is the right way. But the conversion between both functions should be more efficient. Should be confirmed, but I think that the bottleneck is in the conversion, not in the tag detection.

I'm not sure to understand your question about `rgb2gray(unsigned char *rgb, unsigned char *grey, unsigned int size)`. This [function](http://visp-doc.inria.fr/doxygen/visp-daily/classvpImageConvert.html#a4e153df951cd4e8a097c080cc5b594c1) assumes a `rgb` bitmap as input [R_1 G_1 B_1 R_2 G_2 B_2 ... R_n G_n B_n] where `size = n`, meaning the size of the rgb bitmap is equal to `3*size`. It updates as output a grey bitmap that is of size `n` using the formula `grey_i = 0.2126 * R_i + 0.7152 * G_i + 0.0722 * B_i`.

Some ideas:
- The time of your conversion could be compared to the one when using OpenCV [functions](https://github.com/opencv/opencv/blob/master/modules/java/generator/src/cpp/utils.cpp#L23). This is just to confirm what I suspect.
- The conversion should be direct from color `bitmap(in java)` to grey `vpImage<unsigned char>`
- Why not simply trying to introduce a new `argb2grey()` function (not tested):
````
void vpImageConvert::argb2gray(unsigned char *argb, unsigned char *grey, unsigned int size)
{
  unsigned char *pt_input = argb;
  unsigned char *pt_end = argb + size * 4;
  unsigned char *pt_output = grey;
  
  while (pt_input != pt_end) {
    *pt_output = (unsigned char)(0.2126 * (*pt_input+1) + 0.7152 * (*(pt_input + 2)) + 0.0722 * (*(pt_input + 3)));
    pt_input += 4;
    pt_output++;
  }
}
````
- I would be also interested to improve the demo displaying some info in overlay showing that we succeed to localise the tag. I'm thinking about something similar to `vpDisplay::displayLine()` and `vpDisplay::displayFrame()`. There could be a button in the app that activates/disables the display.
- Later when the demo is working faster, create a new `java` module in ViSP that will contain some tools (conversions...) that could be used in every Android app.

AKS1996:
Hi @fspindle 

While working to improve framerate, I read that we can get the live camera feed in YUV `byte[ ]` instead of a bitmap. This means that tag detection can work with just the grey scale data found in the  __Y__ plane of the preview, so its fast now (~10ms or ~15ms for 4 tags). I haven't used `AndroidBitmap_locksPixels` since I'm getting preview as YUV `byte[ ]` not bitmap

And I read that ViSP supports [a host of functions](http://visp-doc.inria.fr/doxygen/visp-daily/classvpImageConvert.html) for conversion from various formats(YUV, YCbCr, etc) to RGBa. We can get an `vpImage<vpRGBa>` from it.

Now I'm trying to implement `vpDisplay::displayLine()` function. I had to ask whether to use opengl or OGRE for a demo(NDK supports opengl by default).

Thanks
Akshay

AKS1996:
Hi all

Recently I was trying to demonstrate a better tag detection using `vpDisplay::displayLine()` or something like a 3d axis on the markers(using pose estimation). Since the functions provided in ViSP depend on OpenCV 3rd party, I tried [doing so without 3rd party](https://github.com/AKS1996/GSOC-Prep/blob/simple_graphics/VispAndroidDemoApp/app/src/main/cpp/native-lib.cpp#L123).

But the images are not well calibrated([get the video here](https://streamable.com/o8aix), resulting lines across markers are skewed). Even to calibrate the camera we need OpenCV.

So **should I try adding OpenCV** support(some changes in the build script) or **start a `visp_java` module** that can be used to tap ViSP functions that don't require any 3rd party.

PR #311, opened 2018-03-20
correcting build error in examples when mbt module is not built

MarcPouliquenInria:
When BUILD_MODULE_visp_mbt is disabled in CMake, a build error occurs in a mbt example. See : https://travis-ci.org/MarcPouliquenInria/ustk/builds/355485512 for example


s-trinh:
Thank you.

PR #332, opened 2018-04-18
Introduce examples using vpMbGenericTracker

fspindle:


fspindle:
We need also to update the corresponding [tutorials](http://visp-doc.inria.fr/doxygen/visp-daily/tutorial_tracking.html)

PR #340, opened 2018-04-24
Add basic template matching

s-trinh:
Add basic template matching as there is now `vpImageTools::normalizedCorrelation()`.
Add corresponding tests.

### DIFF
--- a/modules/core/include/visp3/core/vpImage.h
+++ b/modules/core/include/visp3/core/vpImage.h
@@ -185,6 +185,8 @@ public:
   Type getMinValue() const;
   // Look for the minumum and the maximum value within the bitmap
   void getMinMaxValue(Type &min, Type &max) const;
+  // Look for the minumum and the maximum value within the bitmap and get their location
+  void getMinMaxLoc(vpImagePoint *minLoc, vpImagePoint *maxLoc, Type *minVal=NULL, Type *maxVal=NULL) const;
 
   /*!
 
@@ -221,7 +223,7 @@ public:
   // Gets the value of a pixel at a location with bilinear interpolation.
   Type getValue(vpImagePoint &ip) const;
 
-  // return image sum (for unsigned char and double template types only)
+  // Get image pixels sum
   double getSum() const;
 
   /*!
@@ -631,8 +633,9 @@ template <class Type> void vpImage<Type>::init(unsigned int h, unsigned int w, T
 {
   init(h, w);
 
-  for (unsigned int i = 0; i < npixels; i++)
-    bitmap[i] = value;
+//  for (unsigned int i = 0; i < npixels; i++)
+//    bitmap[i] = value;
+  std::fill(bitmap, bitmap+npixels, value);
 }
 
 /*!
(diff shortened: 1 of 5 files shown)


modules/core/include/visp3/core/vpImage.h
```diff
@@ -185,6 +185,8 @@ template <class Type> class vpImage
   Type getMinValue() const;
   // Look for the minumum and the maximum value within the bitmap
   void getMinMaxValue(Type &min, Type &max) const;
+  // Look for the minumum and the maximum value within the bitmap and get their location
+  void getMinMaxLoc(vpImagePoint *minLoc, vpImagePoint *maxLoc, Type *minVal=NULL, Type *maxVal=NULL) const;
```
fspindle: I would prefer here parameters passed as references rather than pointers

modules/core/include/visp3/core/vpImageTools.h
```diff
@@ -166,6 +171,11 @@ class VISP_EXPORT vpImageTools
   // Linear interpolation
   static float lerp(const float A, const float B, const float t);
 
+  static double normalizedCorrelation2(const vpImage<double> &I1, const vpImage<double> &I2,
+                                       const vpImage<double> &II, const vpImage<double> &IIsq,
```
fspindle: I suggest to rename this function in `normalizedCorrelation()`

modules/core/test/image/testImageTemplateMatching.cpp
```diff
@@ -0,0 +1,392 @@
...
+    vpImageConvert::convert(I, I_double);
+    vpImageConvert::convert(I_tpl, I_tpl_double);
+    I_score.resize(I.getHeight() - height_tpl, I.getWidth() - width_tpl, 0.0);
+
+    for (unsigned int i = 0; i < I.getHeight()-height_tpl; i += step_v) {
+      for (unsigned int j = 0; j < I.getWidth()-width_tpl; j += step_u) {
+        vpRect roi( vpImagePoint(i, j), vpImagePoint(i+height_tpl-1, j+width_tpl-1) );
+        vpImageTools::crop(I_double, roi, I_cur);
+
+        I_score[i][j] = normalizedCorrelation_gold(I_cur, I_tpl_double);
```
fspindle: Instead to implement `normalizedCorrelation_gold()` twice in this test but also in the previous one, why not adding an addition parameter to `normalizedCorrelation(..., bool gold=false)` that allows to use the non optimized version.
And inside this function ensure that if `gold = false`,  `VISP_HAVE_SSE2` is defined, otherwise use the gold version

modules/core/include/visp3/core/vpImage.h
s-trinh: I would like to keep it like this. This allows:
`I.getMinMaxLoc(&min_loc, NULL, &min_val, NULL);`
as usually we want to retrieve only the min/max value and the min/max location.

fspindle: ok. But then please add a sample code in the doc of this function with an example to get all the values, and an other one to retrieve only min & max

PR #359, opened 2018-05-14
Fix: camera intrinsic parameter in the documentation.

aYasutomo:
Hi,

thank you for your great work. 
I'm now using your AprilTag framework in my work well.

I've found a few mistake in your document [here](http://visp-doc.inria.fr/doxygen/visp-daily/classvpCameraParameters.html).

You say "the previous equation si equivalent to the following:", but it doesn't.
The cause of this is misplaced parameters.

From [your citation](http://visp-doc.inria.fr/doxygen/visp-daily/citelist.html#CITEREF_Marchand16a), correct equation is I think as follows:

```math
\begin{matrix}
u \\
v \\
1
\end{matrix} =
\begin{matrix}
p_x & 0   & u_0  \\
0   & p_y & v_0 \\
0   & 0   & 1
\end{matrix}
\begin{matrix}
X_c / Z_c  \\
Y_c / Z_c   \\
1
\end{matrix}
```

This is the first time to send pull request.
So correct me if I'm wrong.

Best regards.

### DIFF
--- a/modules/core/include/visp3/core/vpCameraParameters.h
+++ b/modules/core/include/visp3/core/vpCameraParameters.h
@@ -72,14 +72,14 @@
   1
   \end{array}\right] =
   \left[ \begin{array}{ccc}
-  u_0 & 0   & p_x  \\
-  0   & v_0 & p_y \\
+  p_x & 0   & u_0  \\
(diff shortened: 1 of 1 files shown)


fspindle:
Thanks

PR #360, opened 2018-05-14
some precisions about interpolation methods for vpImage

MarcPouliquenInria:
Added informations in the doc to compare interpolations methods vpImage::getValue() and vpImageTools::interpolate()
Added default parameterr for interpolation type in vpImageTools::interpolate()
Other changes are due to clang-format style

### DIFF
--- a/modules/core/src/image/vpImageTools.cpp
+++ b/modules/core/src/image/vpImageTools.cpp
@@ -490,7 +490,7 @@ void vpImageTools::normalize(vpImage<double> &I)
   Get the interpolated value at a given location.
   \param I : The image to perform intepolation in.
   \param point : The image point.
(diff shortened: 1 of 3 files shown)


modules/core/src/image/vpImageTools.cpp
```diff
@@ -490,12 +490,13 @@ void vpImageTools::normalize(vpImage<double> &I)
   Get the interpolated value at a given location.
   \param I : The image to perform intepolation in.
   \param point : The image point.
-  \param it : The interpolation type (only interpolation with INTERPOLATION_NEAREST and INTERPOLATION_LINEAR
-  are implemented).
+  \param it : The interpolation type (only interpolation with vpImageTools::INTERPOLATION_NEAREST and
```
s-trinh: `\param method`

PR #365, opened 2018-05-25
Discussion: remove rand() call in MBT to make the behaviour deterministic

s-trinh:
Somehow, I would like to discuss about removing the `rand()` calls in MBT.
Each time I forget about these calls. Recently, I spend some time to debug before realizing that the test failures after #354 (and corresponding threshold update: #363, #364) were not due to an issue in the introduced code but rather that the new projection error feature adds a list of `vpMbtDistanceLine`.

Note: I don't know how to properly "fix" the original code, this is just a suggestion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ before_script:
 
 script: 
   - make
-  - ctest
+  - ctest --output-on-failure
 
 notifications:
   email:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -93,6 +93,6 @@ build_script:
   - if "%CMAKE_GENERATOR%"=="MinGW Makefiles" cmake --build . --config %configuration% --target install -- -j2
   - dir C:\projects\visp\build\install
   - dir %VISP_DLL_DIR%
-  - if "%CMAKE_GENERATOR%"=="Visual Studio 14 2015 Win64" ctest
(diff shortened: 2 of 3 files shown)


fspindle:
I agree to suppress the `rand()` call. 

The modification introduced is rough and efficient on ly if P1 and P2 are different from the values you introduce.

An other idea to compute P3, that comes to me would be to 3D rotate P1P2 (`v_tmp1`) by 90° arround P1 along X, or Y or Z. The selected axis could be the one where P1P2 coordinate has the smallest absolute value. P3 becomes than `P1 + vpRotationMatrix*v_tmp1`. Not tested. Should work but will need more lines of code and computation.

s-trinh:
Thanks. We will discuss about the better approach later.

fspindle:
@s-trinh What do you think about this PR ?

- I don't like so much hard coded values, but why not
- An other idea would be to initialize the pseudo-random number generator using `srand(0)`

s-trinh:
@fspindle 

I think the preferable solutions would be:

1. try to implement the method you mentioned above
2. try to see if it is possible to add a `vpUniRand` member variable in the `vpMbTracker` class and pass this object to the `vpMbtDistanceLine::buildFrom(vpPoint &_p1, vpPoint &_p2)` function
3. as last solution, use hardcoded values

`srand(0)` is not good since it will affect globally `rand()`. Solution 2) is better.

fspindle:
Done in PR #871

PR #369, opened 2018-05-29
Change vpColVector and vpRowVector stack(const double &) to stack(double)

oKermorgant:
I found that the current implementation is not robust to self-stacking :

```
vpColVector d(3);
d[0] = 1; d[1] = 2; d[2] = 3;
d.stack(d[0]); // if you are lucky then d = (1, 2, 3, 1)
```
Passing by value if pretty cheap for a _double_ and solves the problem.


fspindle:
Thanks

PR #370, opened 2018-05-31
QR decomposition and inverseTriangular

oKermorgant:
Fabien, 

As discussed, here are 3 new methods for vpMatrix, to perform QR decomposition and inverting a (upper or lower) square triangular matrix.
They are all in vpMatrix_qr and rely on Lapack.

NB - changes about vector stack also appear, I do not know why.

### DIFF
--- a/modules/core/include/visp3/core/vpColVector.h
+++ b/modules/core/include/visp3/core/vpColVector.h
@@ -262,7 +262,7 @@ public:
     vpArray2D<double>::resize(nrows, ncols, flagNullify);
   }
 
-  void stack(const double &d);
+  void stack(double d);
   void stack(const vpColVector &v);
 
   double sum() const;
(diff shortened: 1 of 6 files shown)


oKermorgant:
Cannot reproduce the AppVeyor fail here, I guess it has to do with VISP_HAVE_LAPACK or VISP_HAVE_LAPACK_C subtilities?

fspindle:
It sounds that `dgeqp3_` is not defined in `3rdparty/clapack`. I will check next week when I will be back from Australia

oKermorgant:
There is indeed no dgegp3.c in clapack directory.
If there is a flag to ensure external Lapack can be used then this function may be available only in this case (as in my computer or in the Travis CI apparently).

QRPivot is used in some very particular cases to change a system Ax = b to x = x1 + Z.z in the case of redundant A (ie dim m x n, m < n, rank(A) < m).
SVD can be a fallback (through try QRPivot / catch) even if it involves more computation.
We still need the special flag to tell that dgegp3 is not here (or better, put it in 3rdparty directory!)

Anyway, enjoy the trip in Australia ;)

oKermorgant:
Fabien, 
I have encountered strange memory issues on the call to the Lapack functions. I will resubmit the PR once this is settled.

fspindle:
Ok, let me know. In the mean time, I will try to introduce `dgeqp3_`

Issue #371, opened 2018-06-05
Wrong conversion from vpRotationMatrix to vpThetaUVector for rotations around pi


jchevrie:
The conversion method `vpThetaUVector vpThetaUVector::buildFrom(const vpRotationMatrix &R)` sometimes leads to wrong results when converting a rotation by an angle around pi.

For such rotations, the implemented method is sensitive to the numerical precision on the terms in the rotation matrix.
Some components of the resulting angle/axis representation are then sometimes inverted with respect to the correct values.

PR #372, opened 2018-06-05
Fix wrong conversion from vpRotationMatrix to vpThetaUVector for rotation around pi (#371)

jchevrie:
Fixed the conversion issue in `vpThetaUVector vpThetaUVector::buildFrom(const vpRotationMatrix &R)` for rotation around pi.
New implementation is more numerically stable.


fspindle:
Thanks

PR #375, opened 2018-06-06
QR tools

oKermorgant:
No more memory issues in calling Lapack functions.
- QR and QR Pivot decomposition
- square triangular matrix inverse
- solveByQR (similar but faster than solvebySVD) 

QR Pivot and solveByQR use the dgegp3_ function

The numerous commits are due to my IDE using 4 spaces for indentation, I have to remove them afterwards...

### DIFF
--- a/3rdparty/clapack/src/dgeqp3.c
+++ b/3rdparty/clapack/src/dgeqp3.c
@@ -316,7 +316,7 @@ static integer c__2 = 2;
 
 
             topbmn = minmn - nx;
-/* L30: */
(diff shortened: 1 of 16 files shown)


oKermorgant:
I have set up a virtual machine to test the compilation without GSL or external Lapack, allowing to find which external Lapack c-file were missing. 
The LP / QP solvers will be coming in a future PR.

fspindle:
Up to now nice work ;-)
The LP / QR solvers are welcome

PR #378, opened 2018-06-13
added new tutorial for mbt initialization with aprilTag

MarcPouliquenInria:
New tutorial added for MBT initialization based on aprilTag.
Source code and doxygen tutorial provided.


tutorial/tracking/model-based/generic-aprilTag/tutorial-mb-generic-tracker-aprilTag-init.cpp
```diff
@@ -0,0 +1,325 @@
...
+      d = new vpDisplayX(I);
+#elif defined(VISP_HAVE_GDI)
+      d = new vpDisplayGDI(I);
+#elif defined(VISP_HAVE_OPENCV)
+      d = new vpDisplayOpenCV(I);
+#endif
+    }
+
+    // Prepare MBT
+    vpMbGenericTracker *tracker = new vpMbGenericTracker;
```
s-trinh: I would advice to not use pointer in C++ when it can be avoided (as it seems to be here after a quick glance).

PR #392, opened 2018-07-02
add simple GUI displayLine, displayArrow VpDisplay

AKS1996:


AKS1996:
@fspindle I guess these many display functions are enough for now.
VpDisplay Demo Display

PR #395, opened 2018-07-05
Update core

AKS1996:
+ Remove unused code from parsers
+ Support vector as return types and arguments, to and fro JNI
+ Extend `visp_core` to include more classes necessary for `vpAprilTag` detection


AKS1996:
@fspindle its complete and ready for merging

PR #401, opened 2018-07-16
Turn cmake policy CMP0045, CMP0072, CMP0075 to new

fspindle:
Need to be tested on Jetson TX2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,17 +90,23 @@ endif()
 
 cmake_minimum_required(VERSION 2.8.12.2)
 
+if(POLICY CMP0020)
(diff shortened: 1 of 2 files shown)


fspindle:
Additional check wrt CMP0045 that was not working when turned to NEW (see #328). At that time (April 2018) there was a cmake error when detecting OpenCV.

Configuration, build and test:
- Ok on TX2 with cmake 3.5.1  and OpenCV 3.3.1 (from libopencv-dev)
- Ok on Ubuntu 14.04 with cmake 2.8.12.2  and OpenCV 2.4.8 (from libopencv-dev)

PR #398, opened 2018-07-11
add list primitive converters

AKS1996:

PR #409, opened 2018-07-19
need to fix failing travis build

AKS1996:
@fspindle Travis build for all PRs was failing due to this. Merge this patch


AKS1996:
Yep. the bug was introduced in  #398

PR #410, opened 2018-07-20
Simplify parallel RANSAC pose estimation code

s-trinh:
Surprisingly it can reduce the computation time to enable the multi-threading version on a real case, e.g. with `testKeyPoint-2`:

| | Sequential version  | Parallel version with 4 threads |
| - | ------------- | ------------- |
| Mean | 24.229 ms  | 17.3893 ms  |
| Median | 26.707 ms  | 17.7639 ms  |
| Std | Std: 7.14888  | 2.88802  |


s-trinh:
Note: C++11 thread is used instead.

PR #413, opened 2018-08-03
Extra flags for third party

AKS1996:
Its better to add some flags to avoid 3rd party code instead of manually defining them


AKS1996:
Why's chocolaty so buggy?
> ant (exited 1) - ant not installed. An error occurred during installation

s-trinh:
This looks fine.

PR #434, opened 2018-09-15
change for realsense 2.16

Tianhuanyu:
edit test file for compile

ref: https://github.com/IntelRealSense/librealsense/issues/2324

PR #437, opened 2018-09-20
Fix compatibility with RealSense version before 2.16

s-trinh:
Related #434


fspindle:
Thanks.
Tested with librealsense 2.10.2 on OSX

PR #439, opened 2018-10-02
Eye to hand and pose estimation improvements

fspindle:


modules/vision/src/pose-estimation/vpPose.cpp
```diff
@@ -332,9 +332,9 @@ double vpPose::computeResidual(const vpHomogeneousMatrix &cMo) const
 
     P.track(cMo);
 
-    residual_ += vpMath::sqr(x - P.get_x()) + vpMath::sqr(y - P.get_y());
+    squared_error += vpMath::sqr(x - P.get_x()) + vpMath::sqr(y - P.get_y());
   }
-  return residual_;
+  return (sqrt(squared_error / listP.size()));
```
s-trinh: @fspindle This is quite a radical change in my opinion.
This will break some code where the normalization is made manually and when we threshold this value.

fspindle: @s-trinh I know, but the usual use case is to compare residual between lagrange and dementhon, not to compare to a threshold. Do you think that it could affect your own developments ?

fspindle: @s-trinh I will revert the changes that brake compat with previous releases

PR #475, opened 2018-12-06
Update vpRealSense2.cpp

XuHaiwang:
correct a little mistake.


fspindle:
In realsense 2.9.0 we should use [proccess](https://github.com/IntelRealSense/librealsense/blob/v2.9.0/include/librealsense2/hpp/rs_processing.hpp#L337).

In realsense 2.9.1 we should use [process](https://github.com/IntelRealSense/librealsense/blob/v2.9.1/include/librealsense2/hpp/rs_processing.hpp#L347)

This exactly what ViSP code does
```
#if (RS2_API_VERSION > ((2 * 10000) + (9 * 100) + 0))
    data = align_to->process(data);
#else
    data = align_to->proccess(data);
#endif
```

Could you check the version that you have. Should be in `/usr/local/include/librealsense2/rs.h` the lines similar too:
```
#define RS2_API_MAJOR_VERSION    2
#define RS2_API_MINOR_VERSION    16
#define RS2_API_PATCH_VERSION    5
```

fspindle:
Any update ?

PR #482, opened 2018-12-19
Fix moment alpha

fspindle:
This new PR follows #473.

A little history:
The issue of out of range [-Pi, Pi] in relative mode (#117) was fixed in 64937a5. But
64937a5 also introduced a bug in reference mode computation. Issue #117 was already fixed by the
updated if (signChange) block.

- This PR reverts back alpha reference moment computation using `atan2()` that was saddly removed in  64937a5.
- This PR fixes also the detection of object symetry for small values of 3rd order `mu` setting a very small threshold
- Following results obtained with `testMomentAlpha.cpp` highligths the issue and the new results 

image                                                                                                              | expected | current   | this PR
-------------------------------------------------------------------------------------------------------------------|----------|-----------|----------
arrow+000 | 0        | -0.128108 | -0.128108
arrow+045 | 45       | 44.8881   | 44.8881
arrow+090 | 90      | **_-0.128108_** | 89.8719
arrow+135 | 135      | **_44.8881_**   | 134.888
arrow+180 | 180        | **_-0.128108_** | 179.872
arrow-135 | -135      | **_44.8881_**   | -135.112
arrow-090 | -90      | **_-0.128108_** | -90.1281
arrow-045 | -45       | **_44.8881_**   | -45.1119

PR #487, opened 2018-12-29
Improve moment alpha

fspindle:
Changes introduced in #482 to compute the alpha moment on the arrow images in relative mode didn't allow to compute the alpha moment successfully on a symmetric object like an ellipse.

This new PR fixes the alpha moment computation on an ellipse and introduces a new test on a non symmetric balein object.

The tests achieved on various objects (rectangle, balein, ellipse with xfig, ellipse with gimp, arrow) with different image sizes (320 x 240 for all the previous objects except for the rectangle where the image size was 3200 x 2400) showed that there is no unique threshold to detect if an object is symmetric. I set the default threshold value to 1e-6, but for the ellipse designed with xfig that produces an ellipse with aliasing this threshold needs to be set to 1e-2 to detect its symmetry.

The idea to normalise the 3rd order centered moments by the object surface `m00` was not retained since it doesn't change the fact that the threshold needs to be adapted depending more on the object aliasing than on the object surface.


fspindle:
@xoox This new PR should improve the behavior of `vpMomentAlpha` class

PR #513, opened 2019-01-17
Add SSE code for vpImageTools::imageDifference()

s-trinh:
Fix also `vpImageTools::imageDifferenceAbsolute()`, `abs` was missing.

See also [Notes on SIMD programming](https://github.com/lagadic/visp/wiki/Notes-on-SIMD-programming)


fspindle:
@s-trinh Thanks

PR #514, opened 2019-01-21
Add initUndistortMap() and remap() to speed-up the process to undistort the images

### LINKED ISSUE #490: Optimize vpImageTools::undistort()
This function [vpImageTools::undistort()](https://visp-doc.inria.fr/doxygen/visp-daily/classvpImageTools.html#ac5ce18feb04a1311f6a085d81f125101) sounds very slow compared to OpenCV equivalent implemented in [cv::undistort()](https://docs.opencv.org/3.1.0/da/d54/group__imgproc__transform.html#ga69f2545a8b62a6b0fc2ee060dc30559d)

We should optimize this function. Some ideas:
- precomputing all the transformations; first image will cost, then next images can used precomputed values
- using std::thread and other c++11 features
- using SSE or OpenMP

Note that this function should work also if c++11 is disabled. In that case we could admit to use the existing code

s-trinh:
resolves #490

To undistort multiples images with the same camera intrinsic parameters and the same distortion coefficients:

- call `vpImageTools::initUndistortMap()` once to compute the transformation map
- call `vpImageTools::remap()` for each image with the transformation map in parameter

For `640x480` color images, 6 threads for `vpImageTools::undistort()` and OpenMP, it should speed-up the computation by around **> 30X**.

For `640x480` grayscale images and 6 threads for `vpImageTools::undistort()` and OpenMP, it should speed-up the computation by around **> 8X**.

### DIFF
--- a/modules/core/include/visp3/core/vpImageTools.h
+++ b/modules/core/include/visp3/core/vpImageTools.h
@@ -126,6 +126,10 @@ public:
   static void imageSubtract(const vpImage<unsigned char> &I1, const vpImage<unsigned char> &I2,
                             vpImage<unsigned char> &Ires, const bool saturate = false);
 
+  static void initUndistortMap(const vpCameraParameters &cam, unsigned int width, unsigned int height,
+                               vpArray2D<int> &mapU, vpArray2D<int> &mapV,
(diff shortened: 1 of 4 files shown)


fspindle:
@s-trinh Thanks

PR #520, opened 2019-01-25
Matrices: euclidean norm and condition number fix

romainla:
Propose a fix for the computation of the euclidean norm of matrices and for the condition number of matrices. In both case, the non-square matrices should be handled

You can find below an archive with a program compiled using this fix and a Matlab file. Both programs perform the same computations and give the same results.

[ConditionNumber_solved.zip](https://github.com/lagadic/visp/files/2796440/ConditionNumber_solved.zip)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,7 +5,7 @@
      Copyright (C) 2005 - 2019 by Inria. All rights reserved.
                       https://visp.inria.fr
 
-ViSP 3.2.0 (released January 21, 2019)
+ViSP 3.2.0 (released January 23, 2019)
   - Contributors:
     . This release was possible thanks to contributions from Fabien Spindler,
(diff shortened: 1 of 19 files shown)


fspindle:
I agree changes for `vpMatrix::cond()` to consider only the relevant number of singular values that is equal to the rank (to not consider values set to 0), but I'm confused with your modification in the `vpMatrix::euclideanNorm()`

[Here](http://mathworld.wolfram.com/FrobeniusNorm.html) and [here](https://en.wikipedia.org/wiki/Matrix_norm#Frobenius_norm) the Euclidean norm also called Frobenius norm is the square root of the sum of all the square values of the matrix as it was initially implemented.

romainla:
For whom it may be of interest, [here](https://stats.stackexchange.com/a/229366) is a pretty detailed answer on different use cases for the Frobenius norm. The rest of the discussion also explains some use cases for the induced L2 norm. I hope it can help.

PR #516, opened 2019-01-22
Update vpImage.h

ahfuzhang:
// faster in unsigned char type
// -O3 improve 50%
// not -O3 improve 20%

PR #522, opened 2019-01-28
Fix and improve vpImage::getValue()

### LINKED ISSUE #521: vpImage::getValue(double, double) issues
Some potential issues for `vpImage::getValue(double, double)`:

https://github.com/lagadic/visp/blob/33c67ca010908440d0bf3172fdbdd39d1e673970/modules/core/include/visp3/core/vpImage.h#L1429-L1480

Documentation looks confusing in my opinion:
> Gets the value of a sub-pixel with coordinates (i,j) with bilinear interpolation. If location is out of bounds, then return the value of the closest pixel.

And later
> \exception vpImageException::notInTheImage : If (i,j) is out of the image.

---

The following check looks not needed for me (since `iround = (unsigned int)floor(i)`):
https://github.com/lagadic/visp/blob/33c67ca010908440d0bf3172fdbdd39d1e673970/modules/core/include/visp3/core/vpImage.h#L1467-L1472

---

Invalid read can occur here:
https://github.com/lagadic/visp/blob/33c67ca010908440d0bf3172fdbdd39d1e673970/modules/core/include/visp3/core/vpImage.h#L1477-L1479

For instance with:
```
vpImage<unsigned char> I(3,4);
std::cout << "getValue(2,3): " << I.getValue(2,3) << std::endl;
```

---

https://github.com/lagadic/visp/blob/33c67ca010908440d0bf3172fdbdd39d1e673970/modules/core/include/visp3/core/vpImage.h#L1591

`vpImagePoint` is modified and the code is duplicated.

s-trinh:
Resolves #521 

Continue work of #516

Improvement when using fixed-point arithmetic is around 1.5X but the old code will round the result whereas the new code will cast the result, otherwise around 1.2X when using cast with both methods.

Difference is around 0.5 pixel difference between old and new method.

### DIFF
--- a/modules/core/include/visp3/core/vpImage.h
+++ b/modules/core/include/visp3/core/vpImage.h
@@ -190,19 +190,16 @@ public:
   void getMinMaxLoc(vpImagePoint *minLoc, vpImagePoint *maxLoc, Type *minVal = NULL, Type *maxVal = NULL) const;
 
   /*!
-
     Get the image number of pixels which corresponds to the image
     width multiplied by the image height.
 
     \return The image number of pixels or image size.
-
 
     \sa getWidth(), getHeight()
    */
   inline unsigned int getNumberOfPixel() const { return npixels; }
 
   /*!
-
     Get the number of rows in the image.
 
     \return The image number of rows, or image height.
@@ -210,6 +207,7 @@ public:
     \sa getHeight()
   */
   inline unsigned int getRows() const { return height; }
+
   /*!
     Get the image size.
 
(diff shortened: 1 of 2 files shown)


fspindle:
@s-trinh and @ahfuzhang Thanks a lot
@ahfuzhang I would appreciate if you can fill the copyright transfer [form](https://visp-doc.inria.fr/right-transfert/visp-right-transfer.pdf) and return it signed by email to Fabien.Spindler `_at_` inria.fr (replacing `_at_` with the appropriate character

PR #533, opened 2019-02-11
Add OpenBLAS, Intel MKL and Atlas explicit support

s-trinh:
To be tested, validated, ...

- FindMKL.cmake taken from [here](https://gist.github.com/scivision/5108cf6ab1515f581a84cd9ad1ef72aa)
- [Using BLAS and LAPACK from C/C++ ](http://nicolas.limare.net/pro/notes/2014/10/31_cblas_clapack_lapacke/)
- to be changed: `VISP_HAVE_LAPACK` should mean that we have a vendor BLAS and LAPACK (Atlas, OpenBLAS, MKL, ...) + Netlib for the reference implementation and we should detect which one to properly adapt the code if needed (at least for MKL)


s-trinh:
Update code:

- `VISP_HAVE_LAPACK` means that either Intel MKL, OpenBLAS or Atlas is available
- `VISP_HAVE_LAPACK_BUILT_IN` means that it will build the LAPACK 3rd party

fspindle:
@s-trinh Thanks

PR #535, opened 2019-02-23
Add support of vpImage<vpRGBa> to vpMbGenericTracker

s-trinh:
For `vpMbEdgeMultiTracker`, `vpMbKltMultiTracker` and `vpMbEdgeKltMultiTracker`, add dummy interfaces only for the compilation. Tracking will not work with `vpImage<vpRGBa>` since these classes are deprecated.


fspindle:
@s-trinh Great job. Thanks

PR #543, opened 2019-03-07
Allow passing librealsense dir from CMake command line

s-trinh:
To be able to do: `ccmake ../ViSP-src -DREALSENSE2_DIR=...`

Maybe there is a better way to handle this in CMake?


fspindle:
@s-trinh I just tested on my side and discover that `$ENV(var)` doesn't handle the `-Dvar=...` case. 
What you propose is I think the right way.

I will open an enhancement to introduce similar things in the other `Find<Package>.cmake` files

PR #553, opened 2019-03-22
Update with AprilTag3 code

s-trinh:
New repository for [Apriltag3](https://april.eecs.umich.edu/papers/details.php?name=krogius2019icra): https://github.com/AprilRobotics/apriltags

New additions:
- `TAG_36h10`, `TAG_25h7` and `TAG_36ARTOOLKIT` are deprecated and there is no detector for `TAG_36ARTOOLKIT`
- new families: `TAG_CIRCLE21h7`, `TAG_CIRCLE49h12`, `TAG_CUSTOM48h12`, `TAG_STANDARD41h12`, `TAG_STANDARD52h13`
- new pose estimation method that can return 2 solutions: `HOMOGRAPHY_ORTHOGONAL_ITERATION`
- deprecated: `setAprilTagRefineDecode` and `setAprilTagRefinePose`
- new parameter: `setAprilTagDecodeSharpening`
- for all the computations the tag frame and the camera frame are aligned, but only if `setZAlignedWithCameraAxis` is set that the returned pose will have aligned frames (AprilTag3 computes pose with aligned frames)
- possibility to get the 2nd solutions for the estimated pose (only with `HOMOGRAPHY_ORTHOGONAL_ITERATION` otherwise it will be the same than the 1st solution)
- possibility to get the reprojection error
- detection time is better with single thread and more or less the same with multiple threads
- new library version is 3.0.0 (it follows the original [repository version](https://github.com/AprilRobotics/apriltags/blob/f428cede2386f4b2546f3a6424494eea48339a77/CMakeLists.txt#L15))

There is a current issue with build time that is too long when using `RelWithDebInfo` and `GCC 5.4.0` (Ubuntu 16.04). Can be reproduced on the [original repository](https://github.com/AprilRobotics/apriltags) (there is `CMake` support).
Looks like with `GCC 5.4.0`, `__attribute__ ((optimize("O0")))` has no effect in `RelWithDebInfo` mode. With big tag family, this attribute disables the compiler optimization and avoid long build time. But this seems to not work with `RelWithDebInfo`. With `GCC 8` it is ok.

Will need further tests for Windows platform (MSVC and MinGW). Probably lots of warnings to be fixed + documentation.

Feel free to test, comment the changes or propose some suggestions for the API.


fspindle:
@s-trinh Thanks

Builds without warning on OSX and runs much faster.
Apriltag 2: Mean / Median / Std: 74.3424 ms ; 72.674 ms ; 8.67434 ms
Apriltag 3: Mean / Median / Std: 45.8111 ms ; 45.2585 ms ; 4.078 ms

I will try to find time to test it on windows this week.

fspindle:
Builds without warning with msvc15 and runs also much faster.
Apriltag 2: Mean / Median / Std: 65.463 ms ; 64.9777 ms ; 2.5672 ms
Apriltag 3: Mean / Median / Std: 40.8454 ms ; 41.3576 ms ; 2.71921 ms

fspindle:
@s-trinh Sounds ok to me. I think that we can merge your PR.

Just to not forget, I introduced issue #559 that doesn't relate to this PR.

s-trinh:
@fspindle Ok, there is just one thing I want to check before.

s-trinh:
The PR is finished on my side.

PR #558, opened 2019-04-01
Add pugixml third-party

s-trinh:
Add pugixml as a third-party library to be able to parse and save XML files "natively".


fspindle:
@s-trinh Perfect, thanks. Sounds ok to me. Tested on OSX and Windows. I think that we could merge this PR

s-trinh:
There are still 2 classes that need to be updated:

- `vpXmlParserHomogeneousMatrix`
- `vpXmlParserRectOriented`

s-trinh:
The PR should be complete now.

Probably the documentation and the site should be updated to remove the dependency to `libxml2` for the MBT module (there is still `vpXmlParser` with `libxml2` required but this class is not used in ViSP anymore):

- http://visp.inria.fr/software-architecture/
- https://visp.inria.fr/3rdparty_mbt/
- https://visp.inria.fr/3rd_xml2/

s-trinh:
Data are now saved in full precision.
There should be no more precision loss when saving then reading an XML file and compat.

fspindle:
@s-trinh I propose to remove `vpXmlParser` class, `testXmlParser.cpp` and all the cmake material to detect `libxml2` (`FindICONV.cmake`, `FindXML2.cmake`). Do you agree ?

fspindle:
@s-trinh Finally, I don't think that it's a good idea to remove `vpXmlParser` class since it could potentially break compat with previous releases.

PR #567, opened 2019-04-11
Fix Intel MKL ILP64

s-trinh:
- Use dynamic linking for Intel MKL.
- Detect the architecture to use ILP64 or LP64 MKL interface.


fspindle:
@s-trinh I'll let you merge

PR #572, opened 2019-05-03
Add support for acquiring two infrared images for RealSense SDK2

owninz:


fspindle:
@owninz Thanks for your contribution that will be include in PR #578

Issue #574, opened 2019-05-06
Compatibility problem with PCL > 1.9.1


romainla:
We faced a compilation problem while compiling ViSP with PCL1.9.1 obtained from the Github repository of the Point Cloud Library.

This is due to the fact that PCL1.9.1 requires CMAKE_CXX_FLAGS=-std=c++14  option while ViSP is currently using the c++11 standard.

We found a hack modifying the standard in different CMake files, I created a [patch](https://github.com/lagadic/visp/files/3146791/patchForPCL1_9_1.txt)  which can be used with the command `patch -p1 < patchForPCL1_9_1.txt` 

It did the trick but it is not really pretty though, especially because the variable is called USE_CXX11.

I hope it can be of use for people facing similar problem.

Let me know if you need more information,
Regards

Romain

fspindle:
This issue occurs with the master branch where `CMAKE_CXX_STANDARD = 14` was introduced by default.

It doesn't occur with pcl 1.9.1.

To get pcl 1.9.1 from github you should do:
```
$ git clone https://github.com/PointCloudLibrary/pcl.git
$ git checkout pcl-1.9.1
```
Could be checked using:
```
$ git branch
* (HEAD detached at pcl-1.9.1)
  master
```

In your case I'm pretty sure that you have:
```
$ git branch
  master
```
which indicates that you are on the master branch.

fspindle:
Issues occuring with pcl master branch look like:
```
/home/user/pcl/build/install/include/pcl-1.9/pcl/pcl_config.h:7:4: error: #error PCL requires C++14 or above
   #error PCL requires C++14 or above
```

romainla:
I just checked you are right it seems that I cloned the master branch instead of the pcl-1.9.1 branch as I intended to do. Sorry for this mistake.

fspindle:
Fixed in PR #577

PR #577, opened 2019-05-06
Fix pcl after 1.9.1

fspindle:
- Replace `USE_CXX11=[on|off]` with `USE_CXX_STANDARD=[98|11|14]`. The default value is `11`.
- Since pcl > 1.9.1 cxx standard is set by default to 14 in pcl
- To be able to build ViSP, we set `USE_CXX_STANDARD=14` when pcl 3rd party > 1.9.1


modules/core/include/visp3/core/vpArray2D.h
```diff
@@ -172,7 +172,7 @@ template <class Type> class vpArray2D
   \param c : Array number of columns.
   */
   vpArray2D<Type>(unsigned int r, unsigned int c) :
-  #ifdef VISP_HAVE_CXX11
+  #if defined(VISP_HAVE_CXX11) || defined(VISP_HAVE_CXX14)
```
s-trinh: I am wondering if it is not possible to have something like [`__cplusplus`](https://stackoverflow.com/questions/2324658/how-to-determine-the-version-of-the-c-standard-used-by-the-compiler/7132549#7132549). This way the check could be just:
- `VISP_CXX_VERSION >= 201103L` (C++11)
- `VISP_CXX_VERSION >= 201703L` (C++17)

fspindle: May be possible for linux, but I'm not sure that it works with msvc (see [cxx11.cpp](https://github.com/lagadic/visp/blob/master/cmake/checks/cxx11.cpp))

fspindle: @s-trinh I agree that it would simplify the code and allow to consider all the standard if we have something like:
```
#if (VISP_CXX_STANDARD_VERSION >= VISP_CXX_STANDARD_11)
```
rather than 
```
#if defined(VISP_HAVE_CXX11) || defined(VISP_HAVE_CXX14) || defined(VISP_HAVE_CXX17) || defined(VISP_HAVE_CXX2a) 
```
But the question is which values for `VISP_CXX_STANDARD_VERSION`:
- we cannot use `98`, `11`, `14`, `17`, `2a`
- what do you think about hexa values like: `0x1998`, `0x2011`, `0x2014`, `0x2017`, `0x202a` ?

s-trinh: I would used the same values than `__cplusplus`: `201103L`, `201402L`, `201703L`.

fspindle: Perfect. You are right

fspindle:
Closes #574

PR #583, opened 2019-05-27
Fix pthreads static build

### LINKED ISSUE #559: AprilTag detector freeze with msvc built when more than 1 thread is used
This issue occurs on Windows with msvc14 and msvc15 when `3rdparty/pthreads-win32` is used.

It could be reproduced running:
````
tutorial-apriltag-detector --nthreads 2
````

s-trinh:
fix #559

Using [pthreads4w](https://sourceforge.net/projects/pthreads4w/) solves the issue for me.


fspindle:
@s-trinh Validated with msvc16 and mingw 8.1.0. Thanks

PR #590, opened 2019-06-11
Fix for Visual Studio 12 2013 or previous

fspindle:


fspindle:
Tested with Visual C++ 2010, 2012 and 2015

PR #597, opened 2019-06-25
Fix vpImage bitmap deletion when copyData = false

### LINKED ISSUE #596: vpImage.h SegmentationFault, if bitmap is not copied
Using constructor vpImage with copyData=false, the destructors still deallocates the pointer bitmap. This seems like a bug, since vpImage has no ownership ower the memory for the image. It also results in AccessViolation.

Sample code for testing:
`int main()
{
	{
		cv::Mat img = cv::imread("img.png");
		cv::cvtColor(img, img, cv::COLOR_RGB2GRAY);
		{
			vpImage<unsigned char> I(img.data, img.rows, img.cols, false);
		} // destructor calls delete[]bitmap
	} // crash, since img.data is already freed by vpImage
	return 0;
}`

Suggested simple solution to have a flag, indicating ownership. Later could be improved to a reference counter: 
[vpImage.zip](https://github.com/lagadic/visp/files/3315264/vpImage.zip)

s-trinh:
resolves #596

### DIFF
--- a/modules/core/include/visp3/core/vpImage.h
+++ b/modules/core/include/visp3/core/vpImage.h
@@ -351,6 +351,7 @@ private:
   unsigned int width;   ///! number of columns
   unsigned int height;  ///! number of rows
   Type **row;           ///! points the row pointer array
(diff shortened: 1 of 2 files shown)


fspindle:
Thanks

PR #598, opened 2019-06-26
Add stb_image and stb_image_write to read / write JPEG and PNG

s-trinh:


fspindle:
I was thinking since a while to introduce `libpng` and `libjpeg` as 3rd parties in ViSP, but what you propose is better.

I'm just wondering if we should not move `stb` files in 3rdparty folder. What do you think ?

s-trinh:
Probably it is better to move these files into the 3rd party folder and include them as private headers if necessary. Feel free to make the adequate changes if you have the time.

The main advantage of these header files is the simplicity but compared to `libpng` and `libjpeg`, saved files will be larger (less good compression) and most likely bigger encoding / decoding time.

---

As a side note, it would be interesting to have some feedbacks about how ViSP is used and by whom.
This would help to drive the potential developments:
  - e.g. are the large majority of users on Linux where the dependencies can be installed easily?
  - there are also package managers on Windows like `vcpkg` that allow installing dependencies easily
  - do people prefer having less dependencies, or it is not an issue?
  - running platform (x86 desktop, Raspberry like platform, ...)?
  - etc.

s-trinh:
Maybe a link to a survey in the [http://visp.inria.fr/](http://visp.inria.fr/) page to collect some feedbacks from the community?

fspindle:
Why not like a Google doc page where every one can add/share info. An other idea ?

s-trinh:
I was thinking about an anonymous survey, something like a Google Form. Survey + Google doc page are complementary in my opinion.

For the potential questions:
  - who are you? (student, researcher, developper, freelancer, hobbyist, ... ; which country or region (EU, North America, ...) ; etc.)
  - developping platform (Ubuntu, Windows, MacOS + version, ...)
  - target platform (x86, ARM, Power PC, ...)
  - target hardware (desktop, Raspberry, Jetson, iPhone, Android, ...)
  - which ViSP version are you using?
  - how do you install ViSP (from source, packages, ...)
  - select the (3, 5?) modules you are using the most
  - which functions are you using the most (MBT, template tracker, camera calibration, visual servoing, AprilTag, ...)
  - what other libraries are you using? OpenCV, ROS, Eigen, PCL, YARP, Tensorflow, PyTorch, Keras, other, ...
  - what area to improve? documentation, tutorials, examples, performance, compatibility, ...
  - additional comments?
  - suggestions / comments

fspindle:
I added this nice idea in my todo list ;-)

PR #600, opened 2019-06-30
EPPnP and REPPnP pose computation methods

s-trinh:
Efficient Procrustes PnP (EPPnP) and Robust Efficient Procrustes PnP (REPPnP) C++ code based on the [Matlab implementation code](https://www.iri.upc.edu/people/fmoreno/FMoreno_Publ.html):
```
@inproceedings{Ferraz_cvpr2014,
title = {Very Fast Solution to the PnP Problem with Algebraic Outlier Rejection},
author = {L. Ferraz and X. Binefa and F. Moreno-Noguer},
booktitle = {Proceedings of the Conference on Computer Vision and Pattern Recognition (CVPR)},
pages = {501-508},
year = {2014}
}
```

After implementing the C++ code, unfortunately I did not manage to get good results for the REPPnP method compared to a RANSAC approach on real data. For EPPnP, performance of the approach drops for n <= 6 and the planar case does not handle all the cases in the current implementation.

**Current implemented methods in ViSP are sufficient and I will just summarizes the results here.**

---

Tested methods are:

- Dementhon, Lagrange, Dementhon + VVS and Lagrange VVS from ViSP
- POSIT [from the tutorial](https://visp-doc.inria.fr/doxygen/camera_localization/tutorial-pose-dementhon-visp.html) and POSIT with 3D object coordinates centered
- EPnP from OpenCV

- Rotation error vs number of points with constant Gaussian noise

TestNonPlanar_rotationErrors

- Translation error vs number of points with constant Gaussian noise

TestNonPlanar_translationErrors

- Rotation error vs Gaussian noise

TestNonPlanar_GaussianNoise_rotationErrors

- Translation error vs Gaussian noise

TestNonPlanar_GaussianNoise_translationErrors

- Computation time

TestNonPlanar_computationTime

These figures were plotted using the median and with generated object points centered to zero. With the mean and generated object points skewed (the plots are more scattered when using the mean):

- Rotation error vs Gaussian noise (mean)

TestNonPlanar_GaussianNoise_rotationErrors_skew

- Translation error vs Gaussian noise (mean)

TestNonPlanar_GaussianNoise_translationErrors_skew

---

At the end, Dementhon + VVS will give better results than EPPnP. Since EPPnP is not appropriate when the number of points is <= 6, we cannot use it during the first RANSAC step. Also, since the Dementhon in ViSP "normalizes" the 3D object coordinates, it does not suffer of the center of gravity of the input points.

Lagrange method is very sensitive to the image noise. Also, most of the time we use 4 non-coplanar points to initialize the model-based tracker and the Lagrange method needs at least 6 points.

These are all results from simulated data. A better representation would be to use boxplots and not only the median values.

### DIFF
--- a/modules/core/src/math/matrix/vpMatrix.cpp
+++ b/modules/core/src/math/matrix/vpMatrix.cpp
@@ -1548,12 +1548,7 @@ void vpMatrix::stackColumns(vpColVector &out)
   if ((out.rowNum != colNum * rowNum) || (out.colNum != 1))
     out.resize(colNum * rowNum, false, false);
 
-  double *optr = out.data;
-  for (unsigned int j = 0; j < colNum; j++) {
-    for (unsigned int i = 0; i < rowNum; i++) {
-      *(optr++) = rowPtrs[i][j];
-    }
-  }
+  memcpy(out.data, data, sizeof(double)*out.getRows());
 }
 
 /*!
@@ -1576,11 +1571,7 @@ void vpMatrix::stackRows(vpRowVector &out)
   if ((out.getRows() != 1) || (out.getCols() != colNum * rowNum))
     out.resize(colNum * rowNum, false, false);
 
-  double *mdata = data;
-  double *optr = out.data;
-  for (unsigned int i = 0; i < dsize; i++) {
-    *(optr++) = *(mdata++);
-  }
+  memcpy(out.data, data, sizeof(double)*out.getCols());
 }
 /*!
   Stacks rows of a matrix in a vector.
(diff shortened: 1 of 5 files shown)


s-trinh:
This PR is basically just useless. Here some additional figures with boxplot:

---

- Dementhon vs EPPnP vs EPnP (pose error wrt the number of points):

figure_1

figure_2

- Dementhon vs EPPnP vs EPnP (pose error wrt Gaussian image noise):

figure_1

figure_2

---

- Dementhon VVS vs EPPnP VVS vs EPnP (pose error wrt the number of points):

figure_1

figure_2

- Dementhon VVS vs EPPnP VVS vs EPnP (pose error wrt Gaussian image noise):

figure_1

figure_2

---

- Dementhon VVS vs Lagrange VVS vs EPnP (pose error wrt the number of points):

figure_1

figure_2

- Dementhon VVS vs Lagrange VVS vs EPnP (pose error wrt Gaussian image noise):

figure_1

figure_2

PR #605, opened 2019-07-12
Fix WITH_APRILTAG_BIG_FAMILY

s-trinh:
Fix issue when ViSP is not built with `WITH_APRILTAG_BIG_FAMILY` but a big tag family is selected.

In `RelWithDebInfo` and GCC 5.4.0, build should be ok now.
Not tested on MinGW.
I have only tested that the tutorial code works on VS2017 and release mode.


fspindle:
Thanks

PR #606, opened 2019-07-12
Some various improvements

s-trinh:
`vpRect::getArea()` function is trivial but I use it to compute the intersection over union metric:
```
vpRect bb_inter = bb_true & bb_est;
double IoU = bb_inter.getArea() / (bb_true.getArea() + bb_est.getArea() - bb_inter.getArea());
```

### DIFF
--- a/modules/core/include/visp3/core/vpTime.h
+++ b/modules/core/include/visp3/core/vpTime.h
@@ -43,12 +43,14 @@
 /*!
   \file vpTime.h
   \brief Time management and measurement
-
 */
+#include <visp3/core/vpConfig.h>
 #include <iostream>
 #include <string>
+#if VISP_CXX_STANDARD >= VISP_CXX_STANDARD_11
+#include <chrono>
+#endif
 
-#include <visp3/core/vpConfig.h>
 #include <visp3/core/vpException.h>
 
 /*!
@@ -83,6 +85,26 @@ VISP_EXPORT double measureTimeMicros();
 VISP_EXPORT void sleepMs(double t);
 VISP_EXPORT int wait(double t0, double t);
 VISP_EXPORT void wait(double t);
+};
+
+class VISP_EXPORT vpChrono
+{
+public:
+  vpChrono();
+
+  double getDurationMicros();
+  double getDurationMs();
+  double getDurationSeconds();
+  void start(bool reset=true);
+  void stop();
+
+private:
+  double m_durationMs;
+#if VISP_CXX_STANDARD >= VISP_CXX_STANDARD_11 && (defined(_MSC_VER) && _MSC_VER >= 1900 /* VS2015 */ || !defined(_MSC_VER))
+  std::chrono::steady_clock::time_point m_lastTimePoint;
+#else
+  double m_lastTimePoint;
+#endif
 };
 
 #endif
--- a/modules/core/src/math/matrix/vpMatrix.cpp
+++ b/modules/core/src/math/matrix/vpMatrix.cpp
@@ -1548,12 +1548,7 @@ void vpMatrix::stackColumns(vpColVector &out)
   if ((out.rowNum != colNum * rowNum) || (out.colNum != 1))
     out.resize(colNum * rowNum, false, false);
 
-  double *optr = out.data;
-  for (unsigned int j = 0; j < colNum; j++) {
-    for (unsigned int i = 0; i < rowNum; i++) {
-      *(optr++) = rowPtrs[i][j];
-    }
-  }
+  memcpy(out.data, data, sizeof(double)*out.getRows());
 }
 
 /*!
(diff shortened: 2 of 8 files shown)


modules/core/src/math/matrix/vpMatrix.cpp
```diff
@@ -1639,10 +1630,7 @@ void vpMatrix::kron(const vpMatrix &m1, const vpMatrix &m2, vpMatrix &out)
   unsigned int r2 = m2.getRows();
   unsigned int c2 = m2.getCols();
 
-  if (r1 * r2 != out.rowNum || c1 * c2 != out.colNum) {
-    vpERROR_TRACE("Kronecker prodect bad dimension of output vpMatrix");
-    throw(vpException(vpException::dimensionError, "In Kronecker product bad dimension of output matrix"));
-  }
+  out.resize(r1*r2, c1*c2, false, false);
```
s-trinh: New code will automatically resize the output matrix whereas before the output matrix must be created with the correct size, otherwise it throws an exception.
I don't think it is an issue but in case.

modules/core/src/tools/time/vpTime.cpp
```diff
@@ -80,13 +86,16 @@ double minTimeForUsleepCall = 4;
 double getMinTimeForUsleepCall() { return minTimeForUsleepCall; }
 
 /*!
-  \fn vpTime::measureTimeMs()
-  Return the time in milliseconds since January 1st 1970.
+  Return the time in microseconds since January 1st 1970.
```
s-trinh: Documentation states that it returns the time in microseconds since January 1st 1970.
It should not be true with `QueryPerformanceFrequency`, here [a quick test](https://rextester.com/XKOK1632).

With [`system_clock`](https://en.cppreference.com/w/cpp/chrono/system_clock) it should be better.

modules/core/include/visp3/core/vpTime.h
```diff
@@ -85,4 +87,24 @@ VISP_EXPORT int wait(double t0, double t);
 VISP_EXPORT void wait(double t);
 };
 
+class VISP_EXPORT vpChrono
```
s-trinh: `vpChrono` is to use the C++11 [`steady_clock`](https://en.cppreference.com/w/cpp/chrono/steady_clock), that is a monotonic clock.

Also, documentation of [`gettimeofday`](http://man7.org/linux/man-pages/man2/gettimeofday.2.html):

> The time returned by gettimeofday() is affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the system time).  If you need a monotonically increasing clock, see clock_gettime(2).

Also, by default a call to `chrono.start` reset the chrono but with `chrono.start(false)` it is possible to add the duration.
Other than that, it is redundant with `vpTime::measureTimeMs()`.

modules/core/include/visp3/core/vpTime.h
```diff
@@ -85,4 +87,24 @@ VISP_EXPORT int wait(double t0, double t);
...
+{
+public:
+  vpChrono();
+
+  double getDurationMicros();
```
s-trinh: Not sure about the naming:
 - `getDurationMicroseconds()`?
 - `getDurationMicro()`?
 - `getDurationUs()`?
 - `getDurationMs()`?
 - `getDurationMilli()`?
 - `getDurationMilliseconds()`?
 - ...

I just kept the same naming than for `vpTime::`

[Here](https://en.cppreference.com/w/cpp/chrono/duration) how time is defined in the standard library.

fspindle: Your naming sounds good to me

modules/core/src/math/matrix/vpMatrix.cpp
```diff
@@ -1548,12 +1548,7 @@ void vpMatrix::stackColumns(vpColVector &out)
...
     out.resize(colNum * rowNum, false, false);
 
-  double *optr = out.data;
-  for (unsigned int j = 0; j < colNum; j++) {
-    for (unsigned int i = 0; i < rowNum; i++) {
-      *(optr++) = rowPtrs[i][j];
-    }
-  }
+  memcpy(out.data, data, sizeof(double)*out.getRows());
 }
```
fspindle: I didn't test, but I think that here memcpy cannot be used. We should get column 1, then column 2... while memcpy will stack rows

modules/core/src/math/matrix/vpMatrix.cpp
```diff
@@ -1576,11 +1571,7 @@ void vpMatrix::stackRows(vpRowVector &out)
   if ((out.getRows() != 1) || (out.getCols() != colNum * rowNum))
     out.resize(colNum * rowNum, false, false);
 
-  double *mdata = data;
-  double *optr = out.data;
-  for (unsigned int i = 0; i < dsize; i++) {
-    *(optr++) = *(mdata++);
-  }
+  memcpy(out.data, data, sizeof(double)*out.getCols());
 }
```
fspindle: Yes much better

modules/core/src/math/matrix/vpMatrix.cpp
fspindle: I agree with you

modules/core/src/math/matrix/vpMatrix.cpp
s-trinh: Indeed, my mistakes.

PR #610, opened 2019-07-24
Some various improvements 2

s-trinh:


fspindle:
Do you experienced bad behavior with `lambda = 1` ?

s-trinh:
Initial values are `lambda(0.25), vvsIterMax(200)` in the constructor:
https://github.com/lagadic/visp/blob/87bdede246f2fb80423b6184416f41b01b61e628/modules/vision/src/pose-estimation/vpPose.cpp#L94-L102

The change is to align `init()` with the `vpPose` constructor. Personally, I would remove the `init()` function to prefer constructor.

I have never used the `VVS` with `lambda=1` but I imagine that using a too large gain introduces the risk to diverge maybe?

The change was introduced in #439. If you have experimented that `lambda=1` allows faster convergence and is safe / without the risk to diverge, feel free to dismiss the change and uniform the default values.
But unless I am mistaken, currently this requires to do:
```
vpPose pose;
pose.init();
// or
pose.setLambda(1);
pose.addPoints(points);
```
to actually uses `lambda=1`.
That's why I thought it was a mistake. `vvsIterMax=200` seems also high for `lambda=1` (`vvsEpsilon=1e-8`)?

fspindle:
When I introduced #439 I thought from the doc that changing `lambda = 1` in `init()` was enough to use 1 every where in VVS, but I just discover that `init()` is never called. I remember FC saying that 1. should be used that's why I did the changes in `init()`.

Quick tests with `lambda = 1` are ok. I prefer to not change for the moment `vvsEpsilon=1e-8` and `vvsIterMax=200` even if I agree that this last value could be reduced.

PR #611, opened 2019-08-11
Catch2 for unit and perf tests

s-trinh:
Demo of [Catch2](https://github.com/catchorg/Catch2) for tests and micro-benchmark. Require C++11.

Proposal to use Catch2 to improve tests and add performance benchmark. `ctest` should still work.


s-trinh:
Requires C++11.

For the future, this can be a solution for a simple test framework.

Alternatives are Google Test, doctest, ...

PR #621, opened 2019-09-03
Fix pointcloud computation when align with librealsense2

s-trinh:
Wrong intrinsics / extrinsics were used when align option was used.


s-trinh:
**Not ready yet.**

fspindle:
Thanks. I will merge and we will see later if we can fix that

Issue #661, opened 2019-11-16
Possible issues in image Moments calculations


francofusco:
I believe there is a mistake in the calculation of the invariant `I11` (used to evaluate the invariant `C7` inside `vpMomentCInvariant`). The [wrong line](https://github.com/lagadic/visp/blob/4efcc0b64b9911d676ed3c021d713e21dcfee3a0/modules/core/src/tracking/moments/vpMomentCInvariant.cpp#L137) is:
```c++
I_val[11] = 3 * mu13_2 + 2 * mu31 * mu13 + (-3 * mu40 - 3 * mu04) * mu22 - 2 * mu40 * mu04 + 3 * mu31_2;
```
which, if I am not mistaken, should instead be:
 ```c++
I_val[11] = 3 * mu13_2 + 2 * mu31 * mu13 + (3 * mu40 - 3 * mu04) * mu22 - 2 * mu40 * mu04 + 3 * mu31_2;
```
(the `mu40` inside the brackets should be positive). I checked it against Omar Tahri's PhD thesis and the paper on Image Moments by Omar Thari and François Chaumette and it *should* be as I said - however, please double check it before! :sweat_smile:

If you validate this, can you please back-port the fix to v 3.1?

fspindle:
Thanks for your feedback.
I don't know which is the right equation, since the one in the code corresponds to eq 2.70 in [Manikandan Bakthavatchalam's Phd thesis](http://rainbow-doc.irisa.fr/pdf/2015_these_mani.pdf) page 60.

I will investigate with François Chaumette and let you know.

francofusco:
> I don't know which is the right equation, since the one in the code corresponds to eq 2.70 in Manikandan Bakthavatchalam's Phd thesis page 60.

Thanks for the reply, I will have a look at it!

> I will investigate with François Chaumette and let you know.

Amazing! Meanwhile, I found few other potential issues, I will list them below hoping that they are useful. I apologize in advance for possible mistakes I made below...

---
#### `vpFeatureMomentArea`: possible inconsistency in `get_s` and `interaction`

From what I can understand, the value returned by `vpFeatureMomentArea::get_s` in the case of *discrete* objects is the [sum of the centered moments](https://github.com/lagadic/visp/blob/4efcc0b64b9911d676ed3c021d713e21dcfee3a0/modules/core/src/tracking/moments/vpMomentArea.cpp#L48)

[...]

PR #665, opened 2019-11-20
Minor improvements for image moments

francofusco:
Following the discussion in #661, here there is a small PR that slightly improves `vpFeatureMomentArea` by making sure that the interaction matrix in the discrete case is that of `mu20+mu02` (as in `vpMomentArea`).

Regarding the evaluation of the interaction matrix of Alpha, I saw that the source code already had a [`vpFeatureMomentAlpha::compute_interaction()`](https://github.com/lagadic/visp/blob/0266a1e6c6a25a83bfda4188ab021435b4043e7a/modules/visual_features/src/visual-feature/vpFeatureMomentAlpha.cpp#L49) method that uses the version I was proposing. It is protected by a `#ifdef VISP_MOMENTS_COMBINE_MATRICES` guard. For the moment, I just edited that version. Before going on, I wanted to know what you prefer to do:
- Keep only one definition (the one that is currently ignored)
- Still give the user the possibility to decide which one to use via the `VISP_MOMENTS_COMBINE_MATRICES`, and fixing the second version (as discussed, using `beta=2` and `gamma=1` should solve the problem)

I would probably go for the second one, since I see that many other moments have this feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -448,7 +448,7 @@ VP_OPTION(BUILD_WITH_STATIC_CRT  "" "" "Enables use of staticaly linked CRT for 
 # ----------------------------------------------------------------------------
 VP_OPTION(ENABLE_SOLUTION_FOLDERS "" "" "Solution folder in Visual Studio or in other IDEs" "" (MSVC_IDE OR CMAKE_GENERATOR MATCHES Xcode))
 # Note that it is better to set ENABLE_MOMENTS_COMBINE_MATRICES to OFF
-VP_OPTION(ENABLE_MOMENTS_COMBINE_MATRICES  "" "" "Use linear combination of matrices instead of linear combination of moments to compute interaction matrices." "ENABLE_MOMENTS_COMBINE_MATRICES" OFF)
+VP_OPTION(ENABLE_MOMENTS_COMBINE_MATRICES  "" "" "Use linear combination of matrices instead of linear combination of moments to compute interaction matrices." "" OFF)
 VP_OPTION(ENABLE_TEST_WITHOUT_DISPLAY      "" "" "Don't use display feature when testing" "" ON)
 VP_OPTION(ENABLE_FULL_DOC      "" "" "Build doc with internal classes that are by default not part of the doc" "" OFF)
 
--- a/modules/core/include/visp3/core/vpMomentArea.h
+++ b/modules/core/include/visp3/core/vpMomentArea.h
@@ -50,6 +50,14 @@ class vpMomentCentered; // Required for discrete case of vpMomentObject
 
   \brief Class handling the surface moment.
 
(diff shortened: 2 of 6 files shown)


fspindle:
I would suggest to give the user the possibility to decide which version to use with `ENABLE_MOMENTS_COMBINE_MATRICES` cmake option changing `beta=2` and `gamma=1` in `vpFeatureMomentAlpha.cpp` when discrete.

Moreover, instead to make `ENABLE_MOMENTS_COMBINE_MATRICES` advanced, I suggest also to modify in the main `CMakeLists.txt` file the line:

```
VP_OPTION(ENABLE_MOMENTS_COMBINE_MATRICES  "" "" "Use linear combination of matrices instead of linear combination of moments to compute interaction matrices." "" OFF)
```
so that the user can choose easily during configuration which version to use

francofusco:
I updated `vpFeatureMomentAlpha::compute_interaction()` so that `beta=2`. I removed `gamma` since it is equal to 1 in both cases. I also updated the option in the `CMakeLists.txt` to make it non-advanced as you suggested.

Before making and pushing the commit, however, I wanted to understand one thing. Depending on `VISP_MOMENTS_COMBINE_MATRICES`, two definitions of the class `vpFeatureMomentAlpha` can be generated (see [here](https://github.com/lagadic/visp/blob/0266a1e6c6a25a83bfda4188ab021435b4043e7a/modules/visual_features/include/visp3/visual_features/vpFeatureMomentAlpha.h#L52) and [here](https://github.com/lagadic/visp/blob/0266a1e6c6a25a83bfda4188ab021435b4043e7a/modules/visual_features/include/visp3/visual_features/vpFeatureMomentAlpha.h#L111)). The main difference, aside from the different implementation of `compute_interaction()` is that the method [`error()`](https://github.com/lagadic/visp/blob/0266a1e6c6a25a83bfda4188ab021435b4043e7a/modules/visual_features/include/visp3/visual_features/vpFeatureMomentAlpha.h#L194) is defined only when using the "expanded interaction", *i.e.*, when `ENABLE_MOMENTS_COMBINE_MATRICES` is `OFF`. Is there a specific reason for this? I believe that such distinction is not necessary, and that the `error` method should be the same for both. If this is the case, I would change `vpFeatureMomentAlpha.h` so that it uses a single class definition, to avoid the risk of creating different interfaces. I would also keep the `#ifdef VISP_MOMENTS_COMBINE_MATRICES ... #else ... #end` guard only in `vpFeatureMomentAlpha.cpp` - wrapping just `compute_interaction()` - for a cleaner code. Does it sound good?

fspindle:
I agree, there should be a single class definition. I let you proceed to the changes

francofusco:
I applied the last changes! If CI succeeds, then it should be ready for a merge :+1:

fspindle:
Thanks

Issue #491, opened 2019-01-07
Fix PR #152 to introduce geometric transformations


fspindle:
As stated in PR #152 there is the following bug that should be investigated and fixed before merging PR #152

There is something going wrong when the transformation is set to Id. In that case input image should be the same as warped image.

Introducing some printings shows that the first pixel is not the same in both images:

````
I[0][0]= (81,81,81,255) I[0][0] warped= (88,88,88,255)
````
If you instrument `vpImageTools::warpAffine()` before calling ` getPixelClamped()` with 
````
  std::cout << "imgx[0][0]: " << imgx[0][0] << std::endl;
  std::cout << "imgy[0][0]: " << imgy[0][0] << std::endl;
````
you will get
````
imgx[0][0]: 1
imgy[0][0]: 1
````
meaning that there is a shift of 1 pixel along the lines and the columns

This is confirmed if you print:

````
    std::cout << "I[0][0]= " << I_image[0][0]
        << " I[1][1]= " << I_image[1][1]
        << " I[0][0] warped= "
        << I_warped_image[0][0] << std::endl;
````
you get:
````
I[0][0]= (81,81,81,255) I[1][1]= (88,88,88,255) I[0][0] warped= (88,88,88,255)
````

This shift could explain the left corner being clipped.

Need to be investigated...

PR #152, opened 2017-03-19
Implemented geometric transformation of images

vikasTmz:
Introduced enhancement #146 

@fspindle The remaining pixels around the transformed image now contain random noise after the interpolation. What do I do about it? refer : `line: 501`

Issue #146, opened 2017-03-06
Introduce geometric transformation of images


fspindle:
Today there is noting to rotate or translate an image. Could be useful to introduce these features.

vikasTmz:
Hello @fspindle , I'd like to give this a shot. I've written codes on Affine transformation before so would like to implement this and also understand the code base in the process.

fspindle:
Ok, you are welcome.

vikasTmz:
Can you give me more details on this enhancement?

fspindle:
What I see, but could be discussed is
- output image has the same size than input image
- translation: from a translation vector with 2 doubles as coordinates
- rotation: set the center of rotation (2 double) and the angle in radian
- a scaling factor could also be considered

fspindle:
I better idea would be to implement a more generic function (warpAffine) that takes an affine transformation as input and where each pixel u,v is warped in u_w, v_w using the formula:

u_w = c_u + a_1 * u + a_2*v
v_w = c_v + a_3 * u + a_4*v

In case of a pure translation a_1 = a_2 = a_3 = a_4 = 0
in case of rotation with origin the middle of the image c_u = width/2 and c_v = height/2

vikasTmz:
So, `warpAffine` would take transformation matrix **T** as input?
> u_w = c_u + a_1 * u + a_2v
v_w = c_v + a_3 * u + a_4v

=> **T** = [ a_1 a_3 0 ; 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; a_2 a_4 0 ; 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  c_u c_v 1 ]

I've already implemented a function (_in matlab_) which uses backward transformation (inverse mapping) to warp images, so I know the theory behind this.

Where should this be implemented i.e under which class will `warpAffine` belong to?

fspindle:
``vpImageTools`` class would be a good candicate

fspindle:
Continued in #491

PR #666, opened 2019-11-21
Add warpImage() for affine and perspective image transformations

s-trinh:
supersede #152 
related #146 and #491

Need lagadic/visp-images#10 for CDash.


fspindle:
Thanks

PR #679, opened 2019-12-31
Fix display

s-trinh:
- Fix display of vertical dotted lines with vpDisplayOpenCV
- Fix display with 32-bits X-server (tested on Odroid XU-4)

### DIFF
--- a/modules/gui/src/display/vpDisplayOpenCV.cpp
+++ b/modules/gui/src/display/vpDisplayOpenCV.cpp
@@ -1254,19 +1254,40 @@ void vpDisplayOpenCV::displayCross(const vpImagePoint &ip, unsigned int size, co
   \param color : Line color.
   \param thickness : Line thickness.
 */
-void vpDisplayOpenCV::displayDotLine(const vpImagePoint &ip1, const vpImagePoint &ip2, const vpColor &color,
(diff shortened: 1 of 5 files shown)


modules/gui/src/display/vpDisplayOpenCV.cpp
```diff
@@ -1254,19 +1254,40 @@ void vpDisplayOpenCV::displayCross(const vpImagePoint &ip, unsigned int size, co
...
-      double i = slope * j + orig;
-      displayLine(vpImagePoint(i, j), vpImagePoint(i + deltai, j + deltaj), color, thickness);
+
+    if (vertical_line) {
+      for (unsigned int i = (unsigned int)ip1.get_i(); i < ip2.get_i(); i += (unsigned int)(2 * deltai)) {
+        double j = ip1.get_j();
+        displayLine(vpImagePoint(i, j), vpImagePoint(i + deltai, j), color, thickness);
+      }
+    } else {
+      for (unsigned int j = (unsigned int)ip1.get_j(); j < ip2.get_j(); j += (unsigned int)(2 * deltaj)) {
```
s-trinh: Here `ip1.get_j()` must be < `ip2.get_j()`. So we need to swap  when it is not the case.

PR #689, opened 2020-01-22
Drone ci attempt to print what's happening with testFinMatch, test nu…

fspindle:
…mber 107


fspindle:
To summarize there is something wrong in `testFindMatch.cpp` on drone ci `arm64_gcc_cmake`.
- with commit  d297e36 that have additional debug printings in `vpPoseRansac.cpp`, the test succeed
- while without these printings, there is a segfault. Valgrind doesn't show memleak with suppression files

s-trinh:
My suspect was issue with multithreading.
Maybe [Helgrind](https://valgrind.org/docs/manual/hg-manual.html) would return something useful. The Drone CI has 96 CPUs.

modules/vision/src/pose-estimation/vpPoseRansac.cpp
```diff
@@ -297,8 +297,10 @@ bool vpPose::RansacFunctor::poseRansacImpl()
   }
 
 #if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_11)
-  if (m_nbInliers >= m_ransacNbInlierConsensus)
+  if (m_nbInliers >= m_ransacNbInlierConsensus) {
+    std::lock_guard<std::mutex> guard(m_mutex);
     m_abort = true;
```
s-trinh: @fspindle 
This warning could/should be a false positive. `m_abort` is `std::atomic<bool>` type.
Another possibility is that the OpenBLAS package installed is old (`libopenblas-dev_0.2.20+ds-4_arm64.deb`, ~2017), and maybe there was some data race issue on ARM?

---

Look like the issue appears always on `ARM64 + gcc`, not on 32-bit and neither with clang?
- https://cloud.drone.io/lagadic/visp/276/2/2
- https://cloud.drone.io/lagadic/visp/267/2/2
- https://cloud.drone.io/lagadic/visp/264/2/2
- https://cloud.drone.io/lagadic/visp/289/3/2
- https://cloud.drone.io/lagadic/visp/283/3/2
- https://cloud.drone.io/lagadic/visp/314/3/2
- https://cloud.drone.io/lagadic/visp/298/2/2

---

One idea, I will submit a PR and test:
- add a build without `OpenBLAS` on `ARM64 + gcc` (+ OpenCV?)
- add a build with `OpenBLAS` built from source

fspindle: @s-trinh 
It couldn't be a false positive since the test is failing on arm64.
But I agree that since `m_abort` is atomic, we shouldn't have to introduce the mutex.

The issue can come from `OpenBLAS` (I don't see why) but anyway, we should find a solution to make the test working with every version of `OpenBLAS`. For me the issue comes more from `std::atomic` implementation on `arm64`.

I see 2 options:
- remove `std::atomic<bool> m_abort` var since `m_abort` is not really used in the code (or I missed something)
- replace `std::atomic<bool> m_abort` by `bool m_abort` and use the `mutex`

s-trinh: I believe this is a false positive with atomic:
- [How to avoid false positives with Helgrind?](https://stackoverflow.com/questions/39037964/how-to-avoid-false-positives-with-helgrind)
- [helgrind shows race conditions with atomic types](https://ubuntuforums.org/showthread.php?t=2381299)

This is not specific to a potential implementation issue of `std::atomic` on `arm64`. The Helgrind error can be reproduced with the following code:

<details><summary>code</summary>
<p>

```
#include <thread>
#include <atomic>
#include <vector>
#include <iostream>

namespace
{
class MyThread
{
public:
  MyThread(std::atomic<bool> &abort, int iterations, int id) :
    m_abort(abort), m_iterations(iterations), m_idx(0), m_id(id)
  {
  }

  void operator()() {
    for (m_idx = 0; m_idx < m_iterations && !m_abort; m_idx++) {
      std::this_thread::sleep_for(std::chrono::milliseconds(100));
    }

    m_abort = true;
    //https://ubuntuforums.org/showthread.php?t=2381299&p=13725581#post13725581
//    m_abort.exchange(true);
  }

  int getIterationIndex() const {
    return m_idx;
  }

  int getId() const {
    return m_id;
  }

private:
  std::atomic<bool> &m_abort;
  int m_iterations;
  int m_idx;
  int m_id;
};
}

int main()
{
  std::vector<MyThread> workers;

  std::atomic<bool> abort{false};
  workers.emplace_back(abort, 20, 1);
  workers.emplace_back(abort, 30, 2);
  workers.emplace_back(abort, 10, 0);

  std::vector<std::thread> threadpool;
  for (auto& worker : workers) {
    threadpool.emplace_back(&MyThread::operator(), &worker);
  }

  for (auto& th : threadpool) {
    th.join();
  }

  for (auto &worker : workers) {
    std::cout << "Thread " << worker.getId() << " stopped at iterations: " << worker.getIterationIndex() << std::endl;
  }

  return 0;
}
```

</p>
</details>

---

The idea was to:
- split the N RANSAC iterations equally into M threads
- when one thread found a solution, it sets `m_abort` to `true` to cancel the other threads

Somehow, I forgot to add the check of `m_abort` in the iteration loop.

So I agree to remove `m_abort` and make the necessary changes in the code in another PR for a clean history.

fspindle:
Fix implemented in PR #724

PR #693, opened 2020-01-31
Add vpImageDraw and vpFont

s-trinh:
- Add `vpImageDraw` to draw geometric primitives and avoid the need to use a vpDisplay to draw debug or results information.
 - Add `vpFont` to draw text on an image with the possibility to choose the font size and the background color.

Examples:

image

Tracking results without `vpDisplay`:

image

TODO:
- Fix warnings
- Clean, uniformize data type
- Update test

---

@fspindle 
When are you expecting to release the next ViSP version? 
What about the timeline for shipping the next ViSP version into the Debian package system?


fspindle:
@s-trinh 
- I will do my best to release asap before February 15 which is my deadline
- I think that we could be in time to see the next release in Ubuntu 20.04
- Thinks to be done before releasing
   - Introduce this PR #693
   - Other PR that are in the stack will be considered after the release
   - Check if debian packaging is ok by considering lintian potential warnings
   - Make advanced tests (in mind iOS, Android)
   - Update all the doxygen tutorials 

Any other idea ?

s-trinh:
@fspindle 
I would like to improve the Java binding for MBT (if successful) and I think I have spotted a minor error.
Normally, everything should be finished at the end of this week.

---

Maybe make sure that there is no doxygen warnings.
Probably the other things are to make sure that everything build fine (the different projects, demo, ...) with recent OpenCV version and ViSP ROS.

----

For the Debian packaging, which version will be shipped? The next 3.3 release? Or the 3.2.0-2 that seems to be accepted on [this page](https://tracker.debian.org/pkg/visp)?

PR #701, opened 2020-02-05
Fix build on Alpine and Raspi4

s-trinh:


CMakeLists.txt
```diff
@@ -809,7 +809,7 @@ if(UNIX AND NOT ANDROID)
     file(MAKE_DIRECTORY ${VISP_BINARY_DIR}/doc/man/man1)
     add_custom_command(
       OUTPUT ${VISP_BINARY_DIR}/doc/man/man1/visp-config.1.gz
-      COMMAND ${GZIP} --best -c ${CMAKE_CURRENT_SOURCE_DIR}/doc/man/man1/visp-config.1 > ${VISP_BINARY_DIR}/doc/man/man1/visp-config.1.gz
+      COMMAND ${GZIP} -9 -c ${CMAKE_CURRENT_SOURCE_DIR}/doc/man/man1/visp-config.1 > ${VISP_BINARY_DIR}/doc/man/man1/visp-config.1.gz
```
s-trinh: Somehow on Alpine `gzip` does not have the `--best` alias option.
- Ubuntu:
```
gzip --help
Usage: gzip [OPTION]... [FILE]...
Compress or uncompress FILEs (by default, compress FILES in-place).

Mandatory arguments to long options are mandatory for short options too.

  -c, --stdout      write on standard output, keep original files unchanged
  -d, --decompress  decompress
  -f, --force       force overwrite of output file and compress links
  -h, --help        give this help
  -k, --keep        keep (don't delete) input files
  -l, --list        list compressed file contents
  -L, --license     display software license
  -n, --no-name     do not save or restore the original name and timestamp
  -N, --name        save or restore the original name and timestamp
  -q, --quiet       suppress all warnings
  -r, --recursive   operate recursively on directories
      --rsyncable   make rsync-friendly archive
  -S, --suffix=SUF  use suffix SUF on compressed files
      --synchronous synchronous output (safer if system crashes, but slower)
  -t, --test        test compressed file integrity
  -v, --verbose     verbose mode
  -V, --version     display version number
  -1, --fast        compress faster
  -9, --best        compress better
```

- Alpine:

```
gzip --help
BusyBox v1.31.1 () multi-call binary.

Usage: gzip [-cfkdt123456789] [FILE]...

Compress FILEs (or stdin)

	-1..9	Compression level
	-d	Decompress
	-t	Test file integrity
	-c	Write to stdout
	-f	Force
	-k	Keep input files
```

3rdparty/clapack/f2c/sysdep1.h
```diff
@@ -12,8 +12,10 @@
 #define USE_LARGEFILE
 #ifdef ANDROID
 #define OFF_T off64_t
-#else
+#elif defined(__GLIBC__)
 #define OFF_T __off64_t
+#else
+#define OFF_T off64_t
```
s-trinh: Hopefully the `else` is enough. Otherwise `#elif defined(__GNUC__)`?

modules/core/include/visp3/core/vpNetwork.h
```diff
@@ -45,6 +45,7 @@
 #include <iostream>
 #include <stdio.h>
 #include <string.h>
+#include <sys/select.h>
```
s-trinh: Error was something like:

```
visp/modules/core/include/visp3/core/vpNetwork.h:128:3: error: 'fd_set' does not name a type
  128 |   fd_set readFileDescriptor;
      |   ^~~~~~
```

Should be needed from the [doc](https://linux.die.net/man/3/fd_set). Let see if the build is ok on CI.

modules/core/src/tools/file/vpIoTools.cpp
```diff
@@ -85,7 +85,7 @@
 
 // Detect endianness of the host machine
 // Reference: http://www.boost.org/doc/libs/1_36_0/boost/detail/endian.hpp
-#if defined(__GLIBC__)
+#if defined(__GLIBC__) || (defined(__GNUC__) && defined(__BYTE_ORDER__))
```
s-trinh: Was not able to detect the endianness on Alpine and Raspi4.
Taken from:

https://github.com/rofl0r/musl/blob/d05aaedaabd4f5472c233dbbd1ff4bb9c9c99794/include/endian.h#L10

But just realise that the repo is outdated...

Other solution could be to update the checks [with](https://www.boost.org/doc/libs/1_72_0/boost/predef/other/endian.h).

s-trinh: Wondering if the check would still work with `clang` and Alpine on ARM, [maybe](https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers)?

s-trinh:
There is some weird issue with Drone CI. The build takes too long.

On `arm64_gcc_cmake_opencv — Build and Test`:
```
39/133 Test  #39: testMatrixDeterminant ....................................   Passed  552.39 sec
```

whereas on `arm64_gcc_cmake — Build and Test without OpenCV`:
```
39/133 Test  #39: testMatrixDeterminant ....................................   Passed    4.07 sec
```

I will try to validate first on some ARM embedded platforms before.

s-trinh:
Validated on Raspberry pi 4 and Odroid XU4.

PR #704, opened 2020-02-10
Update Java binding for minimal multi-cameras generic MBT

s-trinh:
I will try to finish it the sooner in order to have it in the next release.


s-trinh:
Passing large array, like pointcloud from Java to JNI and iterating over it in the JNI code seems to be really slow.
Not able to fix it. Nevertheless, minimal multi-camera should work.

PR #705, opened 2020-02-10
Fix error computation in the VVS loop for plane normal

s-trinh:
Not yet tested.


s-trinh:
Really quick test. There seems to be no regression when running pose from RGBD.

PR #707, opened 2020-02-10
Fix pkg-config and visp-config scripts

### LINKED ISSUE #706: visp.pc pkg-config file not conform to man page
- In `visp.pc` there is the full path of the libraries like `/usr/local/lib/libopencv_core.so` instead of using `-L` and `-l` options like `-L/usr/local/lib`and `-lopencv_core` as stated in `pkg-config` man page.

- `visp.pc` with full path doesn't allow to build `robotpkg` `genom3` components

fspindle:
- `pkg-config` and `visp-config` are now using -L<libpath> and -l<libname> instead of libraries full path
- see `pkg-config` man page

- Closes #706


fspindle:
Validated on Jetson TX2 when building `genom3` component

PR #715, opened 2020-02-22
Fix build on FreeBSD with GCC

pkubaj:
FreeBSD doesn't have endian.h, but has sys/endian.h instead.


fspindle:
@pkubaj Thank for pointing this issue.

I'm not completly convinced by your modification since using `__linux__` to exclude FreeBSD from including `endian.h` could lead to side effects.

That's why I did an other change in PR #717 this is more explicit for FreeBSD.

fspindle:
Relates issue #716

PR #732, opened 2020-03-26
Fix pkgconfig script

fspindle:
@s-trinh Could you please test this PR and give me feedback. Thanks


s-trinh:
@fspindle 
Yes it solves the issue with my test. Many thanks.
I will let you know in case of further issue.

PR #743, opened 2020-04-14
Remove deprecated model-based multi tracker classes and tutorials

fspindle:
- `vpMbEdgeMultiTracker`, `vpMbEdgeKltMultiTracker`, `vpMbKltMultiTracker` classes were removed
- corresponding tutorial "Markerless model-based tracking with stereo cameras (deprecated)" also
- to achieve stereo tracking, there is `vpMbGenericTracker` class that is available


fspindle:
@s-trinh I think that it's time to remove these old classes to ease `vpMbGenericTracker` evolution

s-trinh:
@fspindle 

Yes, it is better to remove these classes to not have to maintain them.

I don't think there is somehow that is using these classes.

PR #745, opened 2020-04-22
Support for Realsense T265 model and Kannala Brandt camera model introduction

jnader:
Added Intel RealSense T265 interface with ViSP. This operation required changes in vpCameraParameters class.

Note that, further modifications should be done on vpCameraParameters to include 3 vpCameraParametersProjType.


example/device/framegrabber/getRealSense2Info.cpp
```diff
@@ -50,7 +50,11 @@ int main()
     vpRealSense2 rs;
     rs.open();
     std::cout << "RealSense characteristics:\n" << rs << std::endl;
-    std::cout << "Depth scale: " << std::setprecision(std::numeric_limits<float>::max_digits10) << rs.getDepthScale() << std::endl;
+
+    std::string product = rs.getProductLine();
+
+    if(product.compare("D400") == 0) // If it has depth sensor
```
fspindle, line 56: There is also the SR300 camera that we have in the lab and that has depth info.

If would prefer to change like:
```
if(product.compare("T200") != 0) {
  std::cout << "Depth scale: " << std::setprecision(std::numeric_limits<float>::max_digits10) << rs.getDepthScale() << std::endl;
}
```

example/device/framegrabber/grabT265-all.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+  } catch (const vpException &e) {
+    std::cerr << "RealSense error " << e.what() << std::endl;
+  } catch (const std::exception &e) {
+    std::cerr << e.what() << std::endl;
+  }
+
+  return EXIT_SUCCESS;
+}
+#endif
```
fspindle, line 168: Please add an empty line at the end of this file

example/device/framegrabber/grabT265-image-undistort.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  } catch (const vpException &e) {
+    std::cerr << "RealSense error " << e.what() << std::endl;
+  } catch (const std::exception &e) {
+    std::cerr << e.what() << std::endl;
+  }
+
+  return EXIT_SUCCESS;
+}
+#endif
```
fspindle, line 131: Empty line to add

example/device/framegrabber/grabT265-images-and-odometry-async.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  } catch (const vpException &e) {
+    std::cerr << "RealSense error " << e.what() << std::endl;
+  } catch (const std::exception &e) {
+    std::cerr << e.what() << std::endl;
+  }
+
+  return EXIT_SUCCESS;
+}
+#endif
```
fspindle, line 221: Empty line

example/device/framegrabber/grabT265-images-and-odometry.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  } catch (const vpException &e) {
+    std::cerr << "RealSense error " << e.what() << std::endl;
+  } catch (const std::exception &e) {
+    std::cerr << e.what() << std::endl;
+  }
+
+  return EXIT_SUCCESS;
+}
+#endif
```
fspindle, line 160: empty line

example/device/framegrabber/grabT265-images.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+  } catch (const vpException &e) {
+    std::cerr << "RealSense error " << e.what() << std::endl;
+  } catch (const std::exception &e) {
+    std::cerr << e.what() << std::endl;
+  }
+
+  return EXIT_SUCCESS;
+}
+#endif
```
fspindle, line 107: empty line

example/device/framegrabber/grabT265-odometry.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+  } catch (const vpException &e) {
+    std::cerr << "RealSense error " << e.what() << std::endl;
+  } catch (const std::exception &e) {
+    std::cerr << e.what() << std::endl;
+  }
+
+  return EXIT_SUCCESS;
+}
+#endif
```
fspindle, line 129: empty line

modules/core/include/visp3/core/vpCameraParameters.h
```diff
@@ -333,6 +360,8 @@ class VISP_EXPORT vpCameraParameters
   inline double get_v0() const { return v0; }
   inline double get_kud() const { return kud; }
   inline double get_kdu() const { return kdu; }
+  inline std::vector<double> get_distortion_coefs() const { return m_dist_coefs; }
```
fspindle, line 363: There could be a confusion with the pine hole `kud` and `kdu` coefficients. I suggest to rename `get_distortion_coefs()` into `getKannalaBrandtDistortionCoeficients()`

modules/core/src/camera/vpCameraParameters.cpp
```diff
@@ -118,13 +118,31 @@ vpCameraParameters::vpCameraParameters(double cam_px, double cam_py, double cam_
...
+/*!
+  Constructor for projection with Kannala-Brandt distortion model
+
+  \param cam_px,cam_py : pixel size
+  \param cam_u0,cam_v0 : principal points
+  \param coefficients  : distorsion model coefficients
+
+ */
+vpCameraParameters::vpCameraParameters(double cam_px, double cam_py, double cam_u0,
+                                       double cam_v0, std::vector<double> coefficients)
```
fspindle, line 137: `coefficients` should be passed as a const reference `const std::vector<double> &coefficients`

modules/core/src/camera/vpCameraParameters.cpp
```diff
@@ -267,6 +285,40 @@ void vpCameraParameters::initPersProjWithDistortion(double cam_px, double cam_py
...
+
+/*!
+  Initialization with specific parameters using Kannala-Brandt distortion model
+  \param cam_px,cam_py : The ratio between the focal length and the size of a pixel.
+  \param cam_u0,cam_v0 : Principal points coordinates in pixels.
+  \param coefficients  : Distorsion coefficients.
+*/
+void vpCameraParameters::initProjWithKannalaBrandtDistortion(double cam_px, double cam_py, double cam_u0, double cam_v0,
+                                                      std::vector<double> coefficients)
+{
```
fspindle, line 300: `const std::vector<double> &coefficients`

modules/core/src/camera/vpCameraParameters.cpp
```diff
@@ -267,6 +285,40 @@ void vpCameraParameters::initPersProjWithDistortion(double cam_px, double cam_py
...
+  this->projModel = vpCameraParameters::perspectiveProjWithDistortion;
+
+  this->px = cam_px;
+  this->py = cam_py;
+  this->u0 = cam_u0;
+  this->v0 = cam_v0;
+
+  if (fabs(px) < 1e-6) {
+    vpERROR_TRACE("Camera parameter px = 0");
```
fspindle, line 309: This line could be removed `vpERROR_TRACE("Camera parameter px = 0");`

modules/core/src/camera/vpCameraParameters.cpp
```diff
@@ -267,6 +285,40 @@ void vpCameraParameters::initPersProjWithDistortion(double cam_px, double cam_py
...
+    throw(vpException(vpException::divideByZeroError, "Camera parameter px = 0"));
+  }
+  if (fabs(py) < 1e-6) {
+    vpERROR_TRACE("Camera parameter px = 0");
```
fspindle, line 313: Same, remove `vpERROR_TRACE("Camera parameter px = 0");`

modules/sensor/include/visp3/sensor/vpRealSense2.h
```diff
@@ -62,6 +62,7 @@
   \note Supported devices for Intel® RealSense™ SDK 2.0 (build 2.8.3):
     - Intel® RealSense™ Camera D400-Series (not tested)
     - Intel® RealSense™ Developer Kit SR300 (vpRealSense2 is ok)
+    - Intel® RealSense™ Tracking Camera T265 (under development)
```
fspindle, line 65: Change with:
- Intel® RealSense™ Camera D400-Series
- Intel® RealSense™ Developer Kit SR300
- Intel® RealSense™ Tracking Camera T265

modules/sensor/include/visp3/sensor/vpRealSense2.h
```diff
@@ -350,6 +360,10 @@ class VISP_EXPORT vpRealSense2
   //! pointcloud).
   inline void setMaxZ(const float maxZ) { m_max_Z = maxZ; }
 
+  //! Get odometry data from T265 device and tracker's confidence
+  unsigned int getOdometryData(double &ts, vpHomogeneousMatrix *pose, vpColVector *vel, vpColVector *acc);
```
fspindle, line 364: Time stamp should be the last parameter and passed as NULL pointer:
```
unsigned int getOdometryData(vpHomogeneousMatrix *pose, vpColVector *vel, vpColVector *acc, double *ts = NULL);
```

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -217,6 +217,197 @@ void vpRealSense2::acquire(unsigned char *const data_image, unsigned char *const
...
+    (*vel)[0] = static_cast<double>(pose_data.velocity.x);
+    (*vel)[1] = static_cast<double>(pose_data.velocity.y);
+    (*vel)[2] = static_cast<double>(pose_data.velocity.z);
+    (*vel)[3] = static_cast<double>(pose_data.angular_velocity.x);
+    (*vel)[4] = static_cast<double>(pose_data.angular_velocity.y);
+    (*vel)[5] = static_cast<double>(pose_data.angular_velocity.z);
+  }
+
+    if(acc != NULL)
+  {
```
fspindle, line 305: Indentation

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -217,6 +217,197 @@ void vpRealSense2::acquire(unsigned char *const data_image, unsigned char *const
...
+  \param acc                : Pointer to acceleration vector.
+  \param raw_accel          : Pointer to Accelerometer raw data vector.
+  \param raw_gyro           : Pointer to Gyro raw data vector.
+  \param tracker_confidence : Pose estimation confidence (1: Low, 2: Medium, 3: High)
+  \param ts                 : Timestamp.
+*/
+void vpRealSense2::acquire(vpImage<unsigned char> *left, vpImage<unsigned char> *right, vpHomogeneousMatrix *pose,
+             vpColVector *vel, vpColVector *acc, vpColVector *raw_accel, vpColVector *raw_gyro,
+             unsigned int *tracker_confidence, double *ts)
+{
```
fspindle, line 333: Same remark concerning `tracker_confidence`
Does it make sense to give the user access to `raw_accel` and `raw_gyro` ?
Are `raw_accel` and `raw_gyro` data filtered to produce the content of `acc` ?
If useful we can keep, otherwise remove

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -393,15 +584,16 @@ void vpRealSense2::close() { m_pipe.stop(); }
 /*!
    Return the camera parameters corresponding to a specific stream. This
    function has to be called after open().
-   \param stream : stream for which camera intrinsic parameters are returned.
-   \param type : Indicate if the model should include distorsion parameters or not.
+   \param stream : Stream for which camera intrinsic parameters are returned.
+   \param type   : Indicates if the model should include distorsion parameters or not.
+   \param index  : Index of camera in T265 device, 1: Left  2. Right
```
fspindle, line 589: Maybe here you should complete `index` description with something like. For other cameras don't set this parameter or use -1.

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -423,13 +631,14 @@ vpCameraParameters vpRealSense2::getCameraParameters(const rs2_stream &stream,
 /*!
    Get intrinsic parameters corresponding to the stream. This function has to
    be called after open().
-   \param stream : stream for which the camera intrinsic parameters are returned.
+   \param stream : Stream for which the camera intrinsic parameters are returned.
+   \param index  : Index of the stream.
```
fspindle, line 635: Description of `index` to complete here, and maybe each time this parameter is used in a method

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -812,12 +1021,25 @@ void vpRealSense2::getPointcloud(const rs2::depth_frame &depth_frame, const rs2:
 /*!
    Get the extrinsic transformation from one stream to another. This function
    has to be called after open().
-   \param from, to : streams for which the camera extrinsic parameters are returned.
+   \param from, to   : Streams for which the camera extrinsic parameters are returned.
+   \param from_index : Index of the stream from which we will calculate the transformation
   */
-vpHomogeneousMatrix vpRealSense2::getTransformation(const rs2_stream &from, const rs2_stream &to) const
+vpHomogeneousMatrix vpRealSense2::getTransformation(const rs2_stream &from, const rs2_stream &to, int from_index) const
```
fspindle, line 1027: index description 
completed with: 1: from left to right, 2 from right to left camera for T200 camera models. Otherwise don't use this parameter or set it to -1

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -832,6 +1054,58 @@ vpHomogeneousMatrix vpRealSense2::getTransformation(const rs2_stream &from, cons
...
 
+/*!
+  Get timestamped odometry data from T265 device
+  \param ts    : Timestamp
+  \param pose  : Translation and Orientation of T265 device as vpHomogeneous matrix
+  \param vel   : Linear and angular velocities of T265 device
+  \param acc   : Linear and angular velocities of T265 device
+ */
+unsigned int vpRealSense2::getOdometryData(double &ts, vpHomogeneousMatrix *pose, vpColVector *vel, vpColVector *acc)
+{
```
fspindle, line 1065: see my comment in vpRealSense2.h file

example/device/framegrabber/grabT265-image-undistort.cpp
jnader: I will modify this file and commit changes to vpImageTools to have the undistortFisheye function working.

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
jnader: Yes. RS2_STREAM_ACCEL and RS2_STREAM_GYRO contains linear accelerations and angular velocities respectively. I thought we should have them as extra information.

fspindle: Ok, but it could be nice to investigate in order to see if the values of `raw_accel` and `raw_gyro` are exactly the same than the one in `acc`. If this is the case, please complete method description. The prefix `raw_` is confusing to me.

jnader: Okay. I did some research and found that the stream `RS2_STREAM_ACCEL` provides linear acceleration from IMU at 62.5Hz and `RS2_STREAM_GYRO` provides angular velocities from IMU at 200Hz. But since we are wainting for frames `using wait_for_frames() `, we have the data at 30Hz (images frequency). 

But `acc` and `raw_acc` (now called `imu_acc`) should be different because `acc` is the result of visual inertial odometry and not directly from IMU. I will compare them and decide whether we should keep the new `acquire()` or not.

jnader: `acc` and `imu_acc` contain different data.

example/device/framegrabber/getRealSense2Info.cpp
```diff
@@ -50,7 +50,11 @@ int main()
     vpRealSense2 rs;
     rs.open();
     std::cout << "RealSense characteristics:\n" << rs << std::endl;
-    std::cout << "Depth scale: " << std::setprecision(std::numeric_limits<float>::max_digits10) << rs.getDepthScale() << std::endl;
+
+    std::string product = rs.getProductLine();
+
+    if(product.compare("T200") != 0) // If it has depth sensor
```
s-trinh, line 56: Wondering if there is a way to check if a device has depth support in the `librealsense` library?
Maybe something like this?

https://github.com/IntelRealSense/librealsense/blob/025fccf76803ee6a6e60de9f18ac6193b7ff8597/unit-tests/unit-tests-live.cpp#L5248

modules/core/include/visp3/core/vpCameraParameters.h
```diff
@@ -351,10 +380,12 @@ class VISP_EXPORT vpCameraParameters
...
 
-  double px, py; //!< pixel size
-  double u0, v0; //!<  principal point
-  double kud;    //!< radial distortion (from undistorted to distorted)
-  double kdu;    //!< radial distortion (from distorted to undistorted)
+  double px, py;                       //!< Pixel size
+  double u0, v0;                       //!< Principal point
+  double kud;                          //!< Radial distortion (from undistorted to distorted)
+  double kdu;                          //!< Radial distortion (from distorted to undistorted)
+  bool m_isKannalaBrandt;              //!< Flag set if Kannala-Brandt camera model (lens distortion)
```
s-trinh, line 387: There is the [`vpCameraParametersProjType` enum](https://visp-doc.inria.fr/doxygen/visp-3.3.0/classvpCameraParameters.html#a256a1d8004646b6e22a996a84a42dd2f).

If possible, I think it is better to add this new model into it.

modules/core/include/visp3/core/vpImageTools.h
```diff
@@ -796,6 +799,28 @@ void vpImageTools::undistort(const vpImage<Type> &I, const vpCameraParameters &c
...
+
+  \param cam : Parameters of the camera causing distortion.
+
+  \param undistI : Undistorted output image. The size of this image
+  will be the same than the input image \e I.
+
+  \sa initUndistortMap, remap
+*/
+template<class Type>
+void vpImageTools::undistortFisheye(const vpImage<Type> &I, const vpCameraParameters &cam, vpImage<Type> &undistI)
```
s-trinh, line 815: There is [`undistort()`](https://visp-doc.inria.fr/doxygen/visp-3.3.0/classvpImageTools.html#aed4d783a3c6e85521b1903d3322eb8bc).

I would suggest to modify `undistort()` instead.

modules/core/include/visp3/core/vpImageTools.h
```diff
@@ -796,6 +799,28 @@ void vpImageTools::undistort(const vpImage<Type> &I, const vpCameraParameters &c
...
+{
+  vpArray2D<int> mapU, mapV;
+  vpArray2D<float> mapDu, mapDv;
+
+  initUndistortMap(cam, I.getWidth(), I.getHeight(), mapU, mapV, mapDu, mapDv);
```
s-trinh, line 820: This will do each time `initUndistortMap` + `remap` whereas `initUndistortMap` can be done once for a particular distortion coefficients. This should matter when processing a camera stream.

A solution I see is to update `undistort()` code to remove the old pthread implementation and use instead the combination `initUndistortMap` + `remap`.

To be checked but I don't think the performance between the original `undistort()` and `initUndistortMap` + `remap` should be significant. Otherwise, if it is too much work to adapt `undistort()` with the pthread/non pthread code to handle the two camera models, it should be fine to keep `undistortFisheye()`.

modules/core/include/visp3/core/vpCameraParameters.h
```diff
@@ -229,6 +232,28 @@
 
   An XML parser for camera parameters is also provided in vpXmlParserCamera.
 
+  <b>II. Kannala-Brandt camera model</b>
```
s-trinh, line 235: Probably `vpXmlParserCamera`, `vpPixelMeterConversion` and `vpMeterPixelConversion` should be updated, at least to issue an exception if the Kannala-Brandt model is not supported.

modules/core/src/image/vpImageTools.cpp
```diff
@@ -492,26 +501,65 @@ void vpImageTools::initUndistortMap(const vpCameraParameters &cam, unsigned int
     return;
   }
 
-  float invpx = 1.0f / px;
-  float invpy = 1.0f / py;
-
-  float kud_px2 = kud * invpx * invpx;
-  float kud_py2 = kud * invpy * invpy;
+  float invpx, invpy;
```
s-trinh, line 504: For primitive types, you can declare these variables inside the loop.

example/device/framegrabber/getRealSense2Info.cpp
fspindle: Yes this would be perfect

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
fspindle: We should then rename `raw_gyro` into `gyro_vel`
and create dedicated functions to get these data at the max frequency
```
// Provides linear acceleration from IMU at 62.5Hz.
void getIMUAcceleration(vpColVector &imu_accel) const;
// Provides angular velocities from IMU at 200Hz.
void getGyroVelocities(vpColVector &gyro_vel) const;
```

modules/core/include/visp3/core/vpCameraParameters.h
fspindle: I confirm. You could introduce `KannalaBrandtProjWithDistortion`in the enum

modules/core/include/visp3/core/vpImageTools.h
fspindle: Yes

modules/core/include/visp3/core/vpImageTools.h
fspindle: As you mention, the plan is to call  `initUndistortMap()` + `remap()` in `undistort()` if performances are similar with the old implementation and to add a test to call `initUndistortMap()` only when camera parameter change or when image size change.

modules/core/include/visp3/core/vpCameraParameters.h
fspindle: Work in progress not commited yet

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
jnader: Okay. I will add these 2 functions.

example/device/framegrabber/getRealSense2Info.cpp
jnader: Will check and modify it.

jnader: Done.

modules/core/include/visp3/core/vpCameraParameters.h
jnader: Done.

modules/core/include/visp3/core/vpImageTools.h
s-trinh: Just a suggestion to use function overloading:
  - one `undistort()` function with `vpCameraParameters` as parameters to do `initUndistortMap()` + `remap()`
  - another `undistort()` function with the maps as parameters to just do `remap()` inside

Since `undistort()` is static, probably the solution is to have the image size and the camera parameters variables static inside the `undistort()` function.
I am not fan of using static variables for the following reasons:
  - what if we want to call `undistort()` from different threads (possible data race issue)?
  - what if we want to undistort simultaneously using two camera parameters a sequence of images?

---

Other than that, this is a cool new feature :+1:

jnader: Yesterday, I did the changes related to `vpRealSense2` and `vpCameraParameters` classes. I will be fixing today `undistort()` according to your suggestions.

fspindle:
@jnader I did a couple of changes ;-)

- Harmonize parameters order in `vpRealsense2::acquire()` and `getImuData()` to have every where vel and acc order (`odo_vel`, `odo_acc`, `imu_vel`, `imu_acc`)
- Fix the way to visualize pose frame
- Introduced std::list to display frame trajectory, but finally not used since the is the `std::pair(confidence, points)`
- Keep only one example, make others as tests

What remains to investigate:
- fix a build issue with java bindings: TBD by @fspindle
    ```
    modules/java_bindings_generator/gen/cpp/core.inl.hpp:3080:5: error: ‘vector_double’ was not declared in this scope
     vector_double _retval_ = me->getKannalaBrandtDistortionCoeficients(  );
    ```
- Examples and tests are working on Ubuntu 18.05. We need to test under OSX and Windows: TBD by @fspindle
- Modify the code to be able to change the config and call `open(config)` multiple times if needed: TBD by @fspindle (I have some ideas)

modules/core/include/visp3/core/vpMeterPixelConversion.h
```diff
@@ -240,6 +246,94 @@ class VISP_EXPORT vpMeterPixelConversion
...
+
+    std::vector<double> k = cam.getKannalaBrandtDistortionCoefficients();
+
+    double theta2 = theta*theta, theta3 = theta2*theta, theta4 = theta2*theta2, theta5 = theta4*theta,
+          theta6 = theta3*theta3, theta7 = theta6*theta, theta8 = theta4*theta4, theta9 = theta8*theta;
+
+    double r_d = theta + k[0]*theta3 + k[1]*theta5 + k[2]*theta7 + k[3]*theta9;
+
+    double scale = (r == 0) ? 1.0 : r_d / r;
+
```
fspindle, line 286: Maybe here we will get a warning during build like comparing to zero is not good.
To avoid this future warning:
```
 double scale = (std::fabs(r) < std::numeric_limits<double>::epsilon()) ? 1.0 : r_d / r;
```

modules/core/include/visp3/core/vpMeterPixelConversion.h
```diff
@@ -240,6 +246,94 @@ class VISP_EXPORT vpMeterPixelConversion
...
+    double theta = atan(r);
+
+    std::vector<double> k = cam.getKannalaBrandtDistortionCoefficients();
+
+    double theta2 = theta*theta, theta3 = theta2*theta, theta4 = theta2*theta2, theta5 = theta4*theta,
+          theta6 = theta3*theta3, theta7 = theta6*theta, theta8 = theta4*theta4, theta9 = theta8*theta;
+
+    double r_d = theta + k[0]*theta3 + k[1]*theta5 + k[2]*theta7 + k[3]*theta9;
+
+    double scale = r_d / r;
```
fspindle, line 328: The protection `r != 0` is missing

modules/core/include/visp3/core/vpPixelMeterConversion.h
```diff
@@ -244,6 +250,119 @@ class VISP_EXPORT vpPixelMeterConversion
...
+    double x_d = (u - cam.u0) / cam.px, y_d = (v - cam.v0) / cam.py;
+    double scale = 1.0;
+    double r_d = sqrt(vpMath::sqr(x_d) + vpMath::sqr(y_d));
+
+    r_d = std::min(std::max(-M_PI, r_d), M_PI); // FOV restricted to 180degrees.
+
+    std::vector<double> k = cam.getKannalaBrandtDistortionCoefficients();
+
+    // Use Newton-Raphson method to solve for the angle theta
+    if (r_d > 1e-8)
```
fspindle, line 286: Should use EPS instead of hard coded 1e-8

modules/core/include/visp3/core/vpPixelMeterConversion.h
```diff
@@ -244,6 +250,119 @@ class VISP_EXPORT vpPixelMeterConversion
...
+    {
```
fspindle, line 343: same here

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -641,9 +641,10 @@ void vpRealSense2::acquire(unsigned char *const data_image, unsigned char *const
...
 void vpRealSense2::close()
 {
-  if (m_init) {
-    m_pipe.stop();
-    m_init= false;
+  if (m_pipe != NULL) {
+    m_pipe->stop();
+    delete m_pipe;
+    delete m_pipelineProfile;
   }
```
fspindle, line 648: here you should add `m_pipe = NULL` and `m_pipelineProfile = NULL` after the delete for security

modules/sensor/src/rgb-depth/realsense/vpRealSense2.cpp
```diff
@@ -71,7 +71,7 @@ bool operator==(const rs2_extrinsics &lhs, const rs2_extrinsics &rhs)
  */
 vpRealSense2::vpRealSense2()
   : m_depthScale(0.0f), m_invalidDepthValue(0.0f), m_max_Z(8.0f), m_pipe(), m_pipelineProfile(), m_pointcloud(),
-    m_points(), m_pos(), m_quat(), m_rot(), m_product_line(), m_init(false)
```
fspindle: `m_pipe(NULL), m_pipelineProfile(NULL)`

tutorial/computer-vision/tutorial-pose-from-points-realsense-T265.cpp
```diff
@@ -0,0 +1,159 @@
...
+// Comment / uncomment following lines to use the specific 3rd party compatible with your camera
+//! [Undef grabber]
+//#undef VISP_HAVE_V4L2
+//#undef VISP_HAVE_DC1394
+//#undef VISP_HAVE_CMU1394
+//#undef VISP_HAVE_FLYCAPTURE
+//#undef VISP_HAVE_REALSENSE2
+//#undef VISP_HAVE_OPENCV
+//! [Undef grabber]
+
```
fspindle, line 21: All these commented lines should be removed

PR #751, opened 2020-04-29
Introduce CMP0025 to be able to detect OpenMP with clang compiler

fspindle:
- Propagate OpenMP includes and libraries as well as cflags
- Use `CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES` to detect system headers location for SonarQube instead of hard coded include paths
- Remove workaround for travis that seems thats seems no more necessary


fspindle:
@s-trinh I would be interested to get your feedback concerning errors detected by travis. This PR enables OpenMP usage on OSX that was lost since a while

- There is this [issue](https://travis-ci.org/github/lagadic/visp/jobs/681068626) 
    ```
    error while loading shared libraries: libomp.so
    ```
    That we fixed with this [workaround](https://github.com/lagadic/visp/blob/master/CMakeLists.txt#L537).
    With PR #752 we can see that the missing lib is located in `/usr/local/clang-7.0.0/lib/libomp.so` (see [build](https://travis-ci.org/github/lagadic/visp/jobs/681068626)). Instead of disabling OpenMP on travis + clang, we could modify `LD_LIBRARY_PATH` to help finding `libomp.so`
- Then there are these other errors when running `testPoseRansac2.cpp` (see [build](https://travis-ci.org/github/lagadic/visp/jobs/680934152))
    ```
    There are 968 true inliers found, 1056 inliers returned and 1055 true inliers.
    Problem with the inlier index and the corresponding inlier point!
    
    Returned inliers: oX=-0.015831000099999999, oY=0.1848409999, oZ=-0.020727999899999999 ; x=-0.17276207833729948, y=0.0019804135258031006

    Object points: oX=-0.015831999999999999, oY=0.18484, oZ=-0.020726999999999999 ; x=-0.17276207833729948, y=0.0019804135258031006
    ```
  There is one inlier that is wrong at `1e-6`. I don't see any good fix appart reduscing the threshold to `1e-5`

s-trinh:
@fspindle 

- I found a similar issue [here](https://github.com/travis-ci/travis-ci/issues/8613). Looks like adding the path to `LD_LIBRARY_PATH` would solve the issue. Maybe this is a specific issue to Travis installing `clang` to `/usr/local/clang-7.0.0/bin`? I think that [clang](https://packages.ubuntu.com/bionic/amd64/clang/filelist) on Bionic is installed in `/usr/bin/clang` and [libomp-dev](https://packages.ubuntu.com/bionic/amd64/libomp-dev/filelist) in `/usr/lib/x86_64-linux-gnu/libgomp.so`. Or maybe the `libomp-dev` (LLVM OpenMP runtime) package must be installed?
- This is not normal. The check is to verify that the inlier point and the inlier index correspond to the same point. So this error means that there is somehow an issue with the returned inlier index or inlier point. I will try to have a look at it.

PR #810, opened 2020-08-25
Fix win10 opencv 4.4.0

fspindle:


fspindle:
Relates changes introduced in commit ad84fd8

Issue #806, opened 2020-08-22
Error running the Java sample with intel-mkl 2020.2.254-1


jeremi:
On arch linux, I'm getting an issue running the java tutorial for `ApriltagDetection.java`.
- I'm using the latest `master` ( f42f94d ).
- I have intel-mkl `2020.2.254-1`. My processor is an AMD, but I don't think this is the issue.
- my [ViSP-third-party.txt](https://gist.github.com/jeremi/895cf384c3e09cb31af551813d007f88)

```
TAG_CIRCLE49h12, TAG_CUSTOM48h12, TAG_STANDARD41h12 and TAG_STANDARD52h13 are disabled.
INTEL MKL ERROR: /opt/intel/mkl/lib/intel64/libmkl_def.so: undefined symbol: mkl_sparse_optimize_bsr_trsm_i8.
Intel MKL FATAL ERROR: Cannot load libmkl_def.so.

Process finished with exit code 2
```

fspindle:
Same issue occurs on Ubuntu 18.04 with `l_mkl_2020.0.166.tgz`

1. Note: if Eclipse is not launched from the terminal, running the example from "Run > Run" menu lead to: 

    ````
    Exception in thread "main" java.lang.UnsatisfiedLinkError: /home/fspindle/visp_ws/visp-build/lib/libvisp_java331.so: libmkl_intel_lp64.so: cannot open shared object file: No such file or directory
    ````

2. Starting Eclipse from terminal like `$./eclipse` allows to run the application that opens the interface. But then, after selection of a 36h11 image, pressing `Detect` button lead to the error message:
    ```
    INTEL MKL ERROR: /opt/intel/compilers_and_libraries_2020.0.166/linux/mkl/lib/intel64_lin/libmkl_avx2.so: undefined symbol: mkl_sparse_optimize_bsr_trsm_i8.
    Intel MKL FATAL ERROR: Cannot load libmkl_avx2.so or libmkl_def.so.
    ```

fspindle:
A quick trial shows that this issue is fixed when `cmake/FindMKL.cmake` file is modified to find `mkl_rt` instead of `mkl_core`+`mkl_intel_thread`+`mkl_sequential`

s-trinh:
> I have intel-mkl `2020.2.254-1`. My processor is an AMD, but I don't think this is the issue.

[...]

PR #811, opened 2020-08-25
Add possibility to disable blas/lapack 3rd party usage

fspindle:
- Add OFF to USE_BLAS/LAPACK option list in order to be able to disable lapack 3rd party usage. See #806
- When none of the blas/lapack 3rd party (MKL, OpenBLAS, Atlas, GSL, Netlib) is used, lapack built-in usage is turned ON


fspindle:
Close #806

PR #625, opened 2019-09-18
Introduce support for Kinova Jaco2 robot

fspindle:
@engr-zubair I introduced all the CMake matérial and files so that you can start to modify `vpRobotJaco2` class a wrapper over Jaco2 SDK. This initial work follows PR #619 that you initiate.

PR #839, opened 2020-11-02
Introduce Kinova Jaco interface and examples

fspindle:
This is a fresh PR that tries to cleanup PR #625 removing useless files and synchronizing with visp master branch. 

I have introduced `vpRobotKinova::setDoF()` that allows now to make the `vpRobotKinova` class compatible with 4, 6 or 7 DoF Jaco robot variant.

The examples in `servo-kinova` folder were modified to be able to set the DoF number using `--dof` command line option.

@li2jq I would appreciate if you can test this PR with your 6-DoF Jaco robot

@engr-zubair I would appreciate if you can test this PR with your 7-DoF Jaco robot

Don't hesitate to submit some changes if needed.


engr-zubair:
@fspindle , hi i have tested this code against kinova jaco-2 7 dof, with --dof option, it is working correctly both for cartesian and joint velocities for 7 dof and also if not defined it follows 6 dof behavior as default. 

2. how ever i want to indicate that this visp source code is not building successfully with vc15 2017 when using PCL 1.9.1 third party lib option , giving errors (atth in file) , and also it is only ofering cxx--14 version to use , cannot be build with cxx-11 support . please see the problem , i noticed it is due to some boost libraries which are updated in visp pcl code but are not available in pcl all in one 2017 msvc. 
[visp-3.3-build-vc15-problem.docx](https://github.com/lagadic/visp/files/5558181/visp-3.3-build-vc15-problem.docx)
[visp-3.3-build-vc15-problem.docx](https://github.com/lagadic/visp/files/5558184/visp-3.3-build-vc15-problem.docx)
[visp-3.3-build-vc15-problem.docx](https://github.com/lagadic/visp/files/5558186/visp-3.3-build-vc15-problem.docx)




3. i want  to suggest if we can make the robot.getposition(vprobot::endeffector_frame, pose1) , now here the function returns the pose-1 as vpcolvector , but mostly we need to make a homogenous transform of this position vector , so can we update this in source code that the pose1 comes out as a vpposevector so we can easily use it for making homog.transform from this pose-1.
thanks 
zubair.

fspindle:
@engr-zubair Thanks for your feedback

2/ should be fixed with recent changes in master. PCL need c++14, that's why c++ standard is bumped to c++14 when PCL is used. To keep c++11 you need to siable PCL usage (`USE_PCL=OFF`)

3/ This is not possible, since when joint position is asked we cannot return a `vpPoseVector`. Bellow is the code that shows how to convert a cartesian position returned by `vpRobot::getPosition()` into a `vpHomogeneousMatrix`:

  ```
  vpColVector position;
  robot.getPosition(vpRobot::REFERENCE_FRAME, position);
  vpTranslationVector ftc; // reference frame to camera frame translations
  vpRxyzVector frc; // reference frame to camera frame rotations
  // Update the transformation between reference frame and camera frame
  for (unsigned int i=0; i < 3; i++) {
    ftc[i] = position[i];   // tx, ty, tz
    frc[i] = position[i+3]; // ry, ry, rz
  }
  // Create a rotation matrix from the Rxyz rotation angles
  vpRotationMatrix fRc(frc); // reference frame to camera frame rotation matrix
  // Create fix frame to camera frame transformation in terms of an homogeneous matrix
  vpHomogeneousMatrix fMc(ftc, fRc);
  ```

fspindle:
@engr-zubair I finally introduce a new function `getPosition()` that updates a vpPoseVector. See [documentation](https://visp-doc.inria.fr/doxygen/visp-daily/classvpRobotKinova.html)

PR #845, opened 2020-11-07
Fix ME display when calling setPose() without track() after

### LINKED ISSUE #844: Moving-edges are not updated with vpMbGenericTracker::setPose()
To reproduce this issue launch one of the tutorials in `tutorial/detection/object`.

As shown in the next image, when running `tutorial-detection-object-mbt.cpp` moving-edges location is the one obtained for the first image.

issue-object-detection

s-trinh:
fix #844


fspindle:
Thanks

PR #864, opened 2020-12-09
Simplify code for BgraToRgba conversion in simdlib

s-trinh:
Related #862 

With AVX2, register size is 256 bits. And with RGBa data (offsets are 0, 32, 64, 96), either the whole bitmap is aligned (array address is multiple of 256 bits), either it is not. With RGB data the beginning of the bitmap can be aligned, but never the middle in the computation loop (offsets are 0, 24, 48, 64).
Same thing with SSE that has 128 bits register size.


fspindle:
Ok, thanks.

I must confess that I was not so much confident in my version. But since my tests were passing I thought it was good ;-)

PR #865, opened 2020-12-10
Schedule all Github CI builds every Sunday at 2am

s-trinh:
Reduce scheduled Sanitizers CI build from everyday to every Sunday.
Schedule all the Github CI builds from now.


fspindle:
Thanks

PR #867, opened 2020-12-13
Move most of ViSP SIMD code to simdlib

s-trinh:
The reasons are the following:

- cleaner to have cpp files compiled with only appropriate flags in simdlib (e.g. SSE2 with only `-msse2`, this is not the case if for instance `-march=native` flag is passed for the ViSP code)
- easier to add optimizations for another architectures or for more advanced SIMD instructions set in simdlib
- theoretically, all advanced optimizations would be available in the binary library built from apt:
  - in simdlib, if the compiler supports the generation of AVX2 code, AVX2 code path is available with a runtime check if the CPU supports these instructions set
  - in regular ViSP code, the more advanced flags can only be `-msse2` (or `-mssse3` maybe) on `x86-64` on apt to be compatible with all `x86-64` targets

Remains:
- `vpMbtTukeyEstimator.h`
- `vpMbtFaceDepthDense.cpp`
- `vpMbtFaceDepthNormal.cpp`

### DIFF
--- a/3rdparty/simdlib/Simd/SimdAvx1.h
+++ b/3rdparty/simdlib/Simd/SimdAvx1.h
@@ -1,0 +1,39 @@
+/*
+* Simd Library (http://ermig1979.github.io/Simd).
+*
+* Copyright (c) 2011-2019 Yermalayeu Ihar,
+*               2019-2019 Facundo Galan.
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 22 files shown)


s-trinh:
WIP.

I would like to check something.

s-trinh:
The PR is finished from my side.

fspindle:
@s-trinh Perfect. Thanks

PR #876, opened 2021-01-06
[Test] Reenabling MBT accuracy tests

s-trinh:
Test to see CI results.

Depending on the Math library, not sure if it possible to have accuracy tests since numerical accuracy will depend on the backend for things like SVD, etc.
Maybe increasing the thresholds can be enough if needed.


fspindle:
Ok for me.

s-trinh:
Ok for me also.

PR #882, opened 2021-01-20
Add vpGaussianFilter class for Gaussian blur operation

s-trinh:
It provides similar results with [`scipy.ndimage.gaussian_filter`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html) and `nearest mode` (around 1 pixel difference) and [`Matlab imgaussfilt`](https://www.mathworks.com/help/images/ref/imgaussfilt.html) (less than 1 pixel difference).


fspindle:
Thanks

PR #899, opened 2021-03-09
Fix for packaging 3.4.0 for Debian, Homebrew, ROS Kinetic

fspindle:
- Fix for Debian i386 architecture, see [build log](https://buildd.debian.org/status/fetch.php?pkg=visp&arch=i386&ver=3.4.0-1&stamp=1614944140&file=log)


fspindle:
Packaging sounds ok now with these changes on Debian (appart for armel arch), ROS and homebrew

PR #903, opened 2021-05-26
Re-orthogonalization of rotation part in vpHomogeneousMatrix if slightly not valid

s-trinh:
My use case:
- I am using Blender and poses decimal precision are printed using Blender/Python default terminal accuracy
- I am using C++11 `std::initializer_list` feature to easily initialize a `vpHomogeneousMatrix`:
```cpp
    vpHomogeneousMatrix M1 {
      0.9835, -0.0581,  0.1716, 0.0072,
      -0.0937, -0.9738,  0.2072, 0.0481,
      0.1551, -0.2199, -0.9631, 0.9583
    };
```
- before since the rotation part is not exactly orthogonal, it throws an exception (
- see:
https://github.com/lagadic/visp/blob/efb9421168f7c7770fad1a3ef14020cbcbdf2fbf/modules/core/src/math/transformation/vpHomogeneousMatrix.cpp#L711-L717
https://github.com/lagadic/visp/blob/efb9421168f7c7770fad1a3ef14020cbcbdf2fbf/modules/core/src/math/transformation/vpRotationMatrix.cpp#L399-L436
- now, we perform re-orthogonalization of the rotation part
- so this changes slightly the input data

To be discussed if it is valid to slightly change the input data. Also, this is not updated in `vpRotationMatrix` but the same "issue" exists.

### DIFF
--- a/modules/core/include/visp3/core/vpHomogeneousMatrix.h
+++ b/modules/core/include/visp3/core/vpHomogeneousMatrix.h
@@ -172,7 +172,7 @@ public:
   /*!
     Destructor.
   */
-  virtual ~vpHomogeneousMatrix(){};
+  virtual ~vpHomogeneousMatrix(){}
 
   void buildFrom(const vpTranslationVector &t, const vpRotationMatrix &R);
   void buildFrom(const vpTranslationVector &t, const vpThetaUVector &tu);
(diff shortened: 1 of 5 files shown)


s-trinh:
Can this PR be merged?

I propose to add your suggestions to a feature request.

fspindle:
Sure

s-trinh:
Thanks :+1:

PR #923, opened 2021-09-06
computePlanarObjectPoseFromRGBD function for plane's obstruction case

longhathuc:
As suggestion from Fabien, I have written the functions to compute the pose in case of obstruction. The result is great. Please review if there is any issue. Again, thank you a lot

### DIFF
--- a/modules/vision/src/pose-estimation/vpPoseRGBD.cpp
+++ b/modules/vision/src/pose-estimation/vpPoseRGBD.cpp
@@ -340,6 +340,33 @@ bool vpPose::computePlanarObjectPoseFromRGBD(const vpImage<float> &depthMap, con
   return false;
 }
 
+
+/*!
+  Compute the pose of a planar object from corresponding 2D-3D point coordinates and depth map.
+  Depth map is here used to estimate the 3D plane of the object.
+
+  This implementation reserved for the case when multiple tags is put on the plane. And there is an object 
+    like robot arm obstruct the view and intefere with the plane equation estimation, therefore this function 
+    considers only the 3d point inside the visible tags
+
+  For this case, each polygon is an april tag
+
+  \param[in] depthMap : Depth map aligned to the color image from where \e corners are extracted.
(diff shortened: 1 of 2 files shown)


fspindle:
Please before I can accept and review this PR, can you please fill the [right transfer form](http://visp-doc.inria.fr/right-transfert/visp-right-transfer.pdf) given [here](https://visp.inria.fr/contributing-code/).

modules/vision/src/pose-estimation/vpPoseRGBD.cpp
```diff
@@ -248,86 +252,231 @@ bool validPose(const vpHomogeneousMatrix& cMo) {
...
+  unsigned int nb_points_3d = static_cast<unsigned int>(tag_points_3d.size() / 3);
+
+  if (nb_points_3d > 4) {
+    std::vector<vpPoint> p, q;
+
+    // Plane equation
+    vpColVector plane_equation, centroid;
+    double normalized_weights = 0;
+    estimatePlaneEquationSVD(tag_points_3d, plane_equation, centroid, normalized_weights);
+
```
fspindle, line 438: To adress the more generic case where all the tags don't belong to the same plane, an idea would be to compute one plane for each polygon. Thus `estimatePlaneEquationSVD()` should be called for each polygon.

Can you try to modify the code...

longhathuc: Thanks, I'll modify it and test two solutions carefully.

longhathuc: Ah, actually in my case, we also put the tags in same plane for another object. So should I write another function or add an argument to distinguish the case?

fspindle: In a first attempt, you can add a parameter like `bool coplanar_points` that could be useful to compare results. I hope that for all your use cases the generic case (`coplanar_points = false`) would give good results.

PR #924, opened 2021-09-07
fix CPackDeb on Ubuntu

oKermorgant:
Hi Fabien, 

I tried to generate a `deb` package to ease the installation of ViSP on Ubuntu 20.04. It seems it is not available on the standard repositories.
This PR solves a few issues that were raised:
  - README.txt has been renamed to README.md
  - and above all, the libogre dependency is a bit tedious to handle as the name of the package includes its version. My fix proposes to use the packager's version as the minimum version, but that also makes it the required version (`libogre-2.1-dev` is not superior to `libogre-1.9-dev`, but is considered as another package). I have no idea how to solve this with CPack tools but at least the package is created.


fspindle:
Olivier, thanks for your contrib.

From my understanding the package was named:
- `libogre-dev` for version 1.7.3 until Ubuntu 12.04
- `libogre-1.9-dev` for version 1.9.0 since Ubuntu 14.04

To keep compat with Ubuntu 12.04, maybe we could test the version and do something like:

```
if(OGRE_VERSION VERSION_LESS_EQUAL "1.7.3")
  set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libogre-dev (>=1.7.3)")
else()
  string(REGEX MATCH "[0-9]+\\.[0-9]+" OGRE_VERSION_MINOR ${OGRE_VERSION})  
  set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libogre-${OGRE_VERSION_MINOR}-dev (>=${OGRE_VERSION})")
endif()
```

What do you think ?

oKermorgant:
Hi Fabien, 
Yes that is for sure better to keep backwards compatibility.

fspindle:
I will do the changes ASAP.

PR #932, opened 2021-10-05
Add feature TensortRT on NVIDIA devices

jnader:
Added new tutorial for deep learning object recognition that uses TensorRT inference framework on NVIDIA Jetson devices.
The object detection model currently used is `SSD_Mobilenet` V1 pre-trained on PASCAL VOC dataset.


jnader:
@fspindle I added a description of the tutorial in a seperate page, since I had to describe the parts of the code related to `TensorRT` inference.
At the end, I left a place for a video/image for an example of the object detection that of course could be added to youtube VispTeam channel.

For installing TensorRT and CUDA Toolkit, I redirected the user to the official documentation of NVIDIA because the installation is specific for each platform.

PR #936, opened 2021-10-27
Bayer demosaicing functions

s-trinh:
TODO:
- [X] clean code
- [X] doc
- [x] try to add a test
- ~~[ ] alpha channel handling~~ (alpha channel is left as-is)
- [x] 8-bit support

---

Processing time can be quite expensive. Example to demosaic a `2048x2048` Bayer image, it can take roughly 250/300 ms on a modern computer and the Malvar method without multithreading.

@fspindle 
Any experience with Bayer demosaicing? Had the opportunity to work briefly with a camera allowing to retrieve Bayer data but not anymore. Wondering if it can be useful. Probably using manufacturer SDKs are better.


fspindle:
@s-trinh I'm wondering if these new functions should not be prefixed with `deBayering` instead of  `demosaic`. For me it would be more explicit.

s-trinh:
From the https://en.wikipedia.org/wiki/Demosaicing article:
> A demosaicing (also de-mosaicing, demosaicking or debayering)

Looks like `demosaicing` should be the most commonly used term:
- https://www.mathworks.com/help/images/ref/demosaic.html
- https://www.microsoft.com/en-us/research/publication/high-quality-linear-interpolation-for-demosaicing-of-bayer-patterned-color-images/
- https://docs.baslerweb.com/pgi-feature-set
- https://www.flir.fr/support-center/iis/machine-vision/knowledge-base/how-is-color-processing-performed-on-my-cameras-images/
- https://ieeexplore.ieee.org/abstract/document/1395992

For `debayering`:
- https://www.baslerweb.com/en/sales-support/downloads/document-downloads/debayering-sprint/
- https://multipix.com/wp-content/uploads/2018/05/PGI0116.pdf
- https://ieeexplore.ieee.org/abstract/document/8248995

---

I would stick with `demosaicing` and add something like `(also called "debayering")` in the doc.

fspindle:
@s-trinh Then Ok with `demosaicing` also called `debayering`.

I'm testing live bayer8 conversion with an UEye Camera. It works. I just need to clean the code before committing

fspindle:
@s-trinh OK to merge on my side

s-trinh:
@fspindle 
OK for me also.

PR #955, opened 2021-12-06
Use print() function in both Python 2 and Python 3

cclauss:
Legacy `print` statements are syntax errors in Python 3.  `print()` function works in both Python 2 and Python 3.


fspindle:
Thanks

PR #956, opened 2021-12-06
Delete .travis.yml

cclauss:
It has been several months... https://app.travis-ci.com/github/lagadic/visp/builds


s-trinh:
Before, Travis builds were unlimited for Open Source repo.

The main advantage of Travis compared to the other CI services was the possibility to build on big-endian architecture.
Do you know an online service where we could build and test ViSP on big-endian architecture?

Or the only fallback is to use self-hosted Github actions or maybe QEMU?

s-trinh:
See https://github.com/lagadic/visp/pull/966 for a fallback.

PR #957, opened 2021-12-08
Support for Occipital Structure Core devices

jnader:
Added `vpOccipitalStructure` class in `sensor` module to interface with **Structure Core RGB-D sensors**.
This class is a wrapper over [Occipital Structure SDK](https://structure.io/developers) (current version **0.9**).

There are 2 types of structure core sensors: `Monochrome` and `Color`. The sensor also contains an IMU. For more information concerning the hardware, visit this [page](https://structure.io/structure-core).

Tests implemented:
---------------------------------
- [X] Display depth/visible images,
- [X] Grab and display IMU data,
- [X] Use depth and RGB/Gray images to visualize PCL library.
- [x] Test higher depth resolution (if possible, `1280x960` exists in the SDK)
- [ ] Interface multiple Structure Core sensors.

Tutorials implemented:
---------------------------------
- [X] Grab and save images,
- [X] Generic markerless model-based tracker with RGBD data,
- [x] AprilTag detection.

Improvements:
---------------------------------
- [ ] Hide messages when `open()` is called.


fspindle:
@jnader Very nice work. Thanks.
I introduced some changes to improve/complete mainly the documentation.

I tried to disable the verbosity adding in the `vpOccipitalStructure` constructor
```
  ST::setVisualLoggingVerbosity(-1); // Disable verbosity
  ST::setConsoleLoggingVerbosity(0);
  ST::initializeVisualLogging();
  ST::hideAllVisualLoggingWindows();
```
but without success. Maybe you can try to ask Occipital support.

An other thing that I found on macOS is this error when I quit the binaries using a mouse click:
```
$ ./tutorial-grabber-structure-core
...
Capture session error
```

Can you please try on Ubuntu and check if you have the same error. If you have it, could you please investigate with `valgrind` tool.

jnader:
@fspindle Thanks for your feedback!

So for the error when disconnecting the capture session, it was a missed `break` in the callback function.
There's still a `HOST ERROR` message signaled when disconnecting but it seems it's internal (from the firmware itself).
I'll ask Occipital support for that.

fspindle:
In your last commit I saw a call to `exit(1)`. This should be avoided and replaced by throwing a ViSP fatal error exception

modules/core/src/image/vpImageConvert.cpp
```diff
@@ -281,6 +281,86 @@ vpImageConvert::createDepthHistogram( const vpImage< uint16_t > &src_depth, vpIm
...
+*/
+void
+vpImageConvert::createDepthHistogram( const vpImage< float > &src_depth, vpImage< vpRGBa > &dest_rgba )
+{
+  dest_rgba.resize( src_depth.getHeight(), src_depth.getWidth() );
+  static uint32_t histogram[0x10000];
+  memset( histogram, 0, sizeof( histogram ) );
+
+  for ( unsigned int i = 0; i < src_depth.getSize(); ++i )
+    if(!std::isnan(src_depth.bitmap[i]))
```
s-trinh, line 299:
```suggestion
    if(!vpMath::isNaN(src_depth.bitmap[i]))
```

[std::isnan](https://en.cppreference.com/w/cpp/numeric/math/isnan) is C++11 feature.

modules/core/src/image/vpImageConvert.cpp
```diff
@@ -281,6 +281,86 @@ vpImageConvert::createDepthHistogram( const vpImage< uint16_t > &src_depth, vpIm
...
+ */
+void
+vpImageConvert::createDepthHistogram( const vpImage< float > &src_depth, vpImage< unsigned char > &dest_depth )
+{
+  dest_depth.resize( src_depth.getHeight(), src_depth.getWidth() );
+  static uint32_t histogram2[0x10000];
+  memset( histogram2, 0, sizeof( histogram2 ) );
+
+  for ( unsigned int i = 0; i < src_depth.getSize(); ++i )
+    if(!std::isnan(src_depth.bitmap[i]))
```
s-trinh, line 340:
```suggestion
    if(!vpMath::isNaN(src_depth.bitmap[i]))
```

Same.

modules/core/include/visp3/core/vpImageConvert.h
```diff
@@ -109,6 +109,9 @@ class VISP_EXPORT vpImageConvert
   static void createDepthHistogram( const vpImage< uint16_t > &src_depth, vpImage< vpRGBa > &dest_rgba );
   static void createDepthHistogram( const vpImage< uint16_t > &src_depth, vpImage< unsigned char > &dest_depth );
 
+  static void createDepthHistogram( const vpImage< float > &src_depth, vpImage< vpRGBa > &dest_depth );
+  static void createDepthHistogram( const vpImage< float > &src_depth, vpImage< unsigned char > &dest_depth );
```
s-trinh, line 113: Wondering if it is a good idea to add these overloads for `float` type?

Looks like the only difference with `uint16_t` is the handling of `NaN` for incorrect/not available depth.
I don't know if it would be possible to have a `createDepthHistogram` for `float` type but with a very different behaviour?
Maybe renaming the function to something specific to Occipital?

Or split the function behaviour between detecting `NaN`, replacing this value to `0`, converting the `vpImage` to `uint16_t` and calling the `createDepthHistogram` with `uint16_t` type?

s-trinh (commented):
Some comments from a very quick look at the PR.

modules/core/include/visp3/core/vpImageConvert.h
jnader: Thanks @s-trinh for these comments!

In fact, together with @fspindle, we had a similar discussion before implementing these overloads.
Another option was to create a `convert` function from `float` to `uint16_t` that deals with `NaN`. But this `convert` will be just used for creating depth histogram (since depth is returned as float* from the SDK, and we just need to transform this information to a displayed image `vpRGBa/unsigned char`). 

So should we be creating a temporary `uint16_t` image and pass it to the already existing `createDepthHistogram` or create a `createDepthHistogram` for `float` images?

modules/core/src/image/vpImageConvert.cpp
jnader: Done.

modules/core/src/image/vpImageConvert.cpp
jnader: Done.

modules/core/include/visp3/core/vpImageConvert.h
fspindle: For me it is not an issue to have an overloaded function for float image. I prefer this option than introducing a conversion from `float` to `uint16_t` images that is computationnaly costly.

To avoid code duplication, maybe we could create a template function that does the job and that is called in `vpImageConvert::createDepthHistogram()`

I suggest to create the template in `src/image/private/vpImageConvert_impl.h` with something like (not tested)

```
/*!
  Convert the input float depth image to a 8-bits depth image. The input
  depth value is assigned a value proportional to its frequency.
  \param[in] src_depth : Input depth image.
  \param[out] dest_depth : Output grayscale depth image.
  \param[in] check_isNaN : When true test if input depth image values are NaN. 
  If this is the case these pixels are set to 0 in the output image.
 */
template <class Type> 
void
vp_createDepthHistogram( const vpImage< Type > &src_depth, vpImage< unsigned char > &dest_depth, bool check_isNaN )
{
  dest_depth.resize( src_depth.getHeight(), src_depth.getWidth() );
  uint32_t histogram[0x10000];
  memset( histogram, 0, sizeof( histogram ) );

  if (check_isNaN) {
    for ( unsigned int i = 0; i < src_depth.getSize(); ++i ) {
      if(!vpMath::isNaN(src_depth.bitmap[i])) {
        ++histogram[static_cast<uint32_t>(src_depth.bitmap[i])];
      }
    }
  }
  else {
    for ( unsigned int i = 0; i < src_depth.getSize(); ++i ) {
      ++histogram[static_cast<uint32_t>(src_depth.bitmap[i])];
    }
  }

  for ( int i = 2; i < 0x10000; ++i )
    histogram[i] += histogram[i - 1]; // Build a cumulative histogram for the indices in [1,0xFFFF]

  for ( unsigned int i = 0; i < src_depth.getSize(); ++i )
  {
    uint16_t d = src_depth.bitmap[i];
    if ( d )
    {
      unsigned char f =
          static_cast< unsigned char >( histogram[d] * 255 / histogram[0xFFFF] ); // 0-255 based on histogram location
      dest_depth.bitmap[i] = f;
    }
    else
    {
      dest_depth.bitmap[i] = 0;
    }
  }
}
```
where I suppressed the `static` and change the `cast` wrt to current code.

Then in `vpImageConvert.cpp`the code becomes:
```
void
vpImageConvert::createDepthHistogram( const vpImage< float > &src_depth, vpImage< unsigned char > &dest_depth)
{
  vp_createDepthHistogram(src_depth, dest_depth, true);
}

void
vpImageConvert::createDepthHistogram( const vpImage< uint16_t > &src_depth, vpImage< unsigned char > &dest_depth )
{
  vp_createDepthHistogram(src_depth, dest_depth, false);
}
```

Similar thing could be done for `vpRGBa` images.

@s-trinh and @jnader What do you think ?

fspindle: I was just thinking that adding `openmp` support before the `for` iterators could boost the behavior.

jnader: @fspindle Indeed a template function could be a good solution for avoiding code duplication.

jnader: I implemented the private template function and ran some performance tests on `createDepthHistogram()` with and without `openmp` support, on both `RealSense D435` and `Structure Core` sensors, and got the following results:

| Hardware | Without OpenMP<br>(Average execution time) | With OpenMP<br>(Average execution time)|
| ------------- | ------------- | ------------- |
| RealSense D435 | 1.64825 ms | 1.34228 ms |
| Structure Core | 2.92123 ms | 0.776386 ms |

fspindle:
@jnader There are somme build issues around OpenMP usage on WIndows.

jnader:
@fspindle I modified the tutorial for grabbing images from Structure Core.

It now has some extra options for:
1) Grabbing `1280x960` depth arrays,
2) Grabbing depth and visible frames at different fps (no frame synchronization).

Concerning multiple structure core interface, I'm having some issues with the SDK when I want to create `captureSession` with specified `serial number`. I'm still working on it for now.

s-trinh:
Maybe it could be interesting to add some videos on the ViSPTeam YouTube channel showing/comparing the depth images between Intel RealSense and the Structure Core.

And adding the link in the tutorial.

fspindle:
The build issues are related to `vpMath::isNaN()` overloaded function for `float` (see [link](https://github.com/lagadic/visp/runs/4559386634?check_suite_focus=true))

I think that the solution is to modify `vpMath.cpp` introducing a new overloaded function for `short`. The code is for me similar to the one for `float`and could be:

```
/*!
   Check whether a short number is not a number (NaN) or not.
   \param value : Short number to check.
   \return true if value is Not A Number (as defined by IEEE754 standard) and false otherwise.
 */
bool vpMath::isNaN(short value)
{
#if defined(VISP_HAVE_FUNC_ISNAN)
  return isnan(value);
#elif defined(VISP_HAVE_FUNC_STD_ISNAN)
  return std::isnan(value);
#elif defined(VISP_HAVE_FUNC__ISNAN)
  return (_isnan(value) != 0);
#else
  // Taken from OpenCV source code CvIsNan()
  Cv64suf ieee754;
  ieee754.f = value;
  return (((unsigned)(ieee754.u >> 32) & 0x7fffffff) + ((unsigned)ieee754.u != 0) > 0x7ff00000) != 0;
#endif
}
```

s-trinh:
`NaN` in C++ is only defined for floating-point values: https://en.cppreference.com/w/cpp/types/numeric_limits/has_quiet_NaN

Not sure how to solve it.

fspindle:
The only solution I see is to revert to the original version without using a template with a call to `vpMath::isNaN()` only for the version dedicated to `vpImage<float>`

jnader:
@fspindle Exactly! Otherwise, we'll be creating a `isNaN` function for integer values that just returns `false` which is trivial.

@s-trinh Concerning the video, it's an interesting idea, I'll be working on that and add the link to the tutorial.

PR #965, opened 2021-12-23
Implement makeTempDirectory() for Windows

### LINKED ISSUE #948: makeTempDirectory() for Windows
There is [makeTempDirectory()](https://visp-doc.inria.fr/doxygen/visp-3.3.0/classvpIoTools.html#ad2c58abadddb9550268ad6c81fcc62d9) but only for Unix right now.

Could be interesting to update it for Windows also. Maybe using a combination of?
- https://en.wikipedia.org/wiki/Temporary_folder
- [getenv()](https://visp-doc.inria.fr/doxygen/visp-3.3.0/classvpIoTools.html#aa1e1bd46aaedead314c61e3a4d6cde99)

s-trinh:
fix #948

It uses `UuidCreate()` to have something more or less unique. Maybe we should add a check to verify that the folder does not already exist, and if so generate another uuid?

Maybe we should add also `!defined(WINRT)` at some appropriate locations.

### DIFF
--- a/modules/core/src/tools/file/vpIoTools.cpp
+++ b/modules/core/src/tools/file/vpIoTools.cpp
@@ -142,8 +142,54 @@ const std::string &vpIoTools::getBuildInformation()
 
 /*!
   Return path to the default temporary folder:
-    - on Windows it returns GetTempPath()
-    - on Unix it returns /tmp/<username>
+    - on Windows it returns `GetTempPath()`
+    - on Unix it returns `/tmp/<username>`
+
+  \warning This function is not implemented on WINRT.
+
+  The following sample shows how to use this function to create unique temporary directories:
+  \code
+  include <visp3/core/vpIoTools.h>
+
+  int main()
+  {
+    std::string tmp_path = vpIoTools::getTempPath();
+    std::cout << "Temp path: " << tmp_path << std::endl;
+
+    std::string tmp_dir1 = vpIoTools::makeTempDirectory(tmp_path);
+    std::cout << "Created unique temp dir1: " << tmp_dir1 << std::endl;
+
+    std::string tmp_dir2_template = tmp_path + vpIoTools::path("/") + "dir_XXXXXX";
+    std::string tmp_dir2 = vpIoTools::makeTempDirectory(tmp_dir2_template);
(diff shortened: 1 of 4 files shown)


s-trinh:
There is an issue on MacOS 11: https://github.com/lagadic/visp/runs/4644601865?check_suite_focus=true

fspindle:
@s-trinh Thanks

1/ From what I understand from the [UUID doc](https://docs.microsoft.com/en-us/windows/win32/rpc/rpcdce/ns-rpcdce-uuid) there is no need to verify that the folder does not already exist. It should be unique.

2/ Protections against WINRT are added

3/ I know that sometimes build on MacOS 11 is failing. I don't understand why. Il will investigate these days.

4/ If possible, we should add WINRT build to appveyor CI based on this [tutorial](https://visp-doc.inria.fr/doxygen/visp-daily/tutorial-install-win10-msvc16-uwp.html). I don't know how to introduce it properly right now, but I will check this point...

For me, this PR could be merged

s-trinh:
For 3/ somehow this looks like an issue with SIFT for me:

> You should not use KAZE with Pyramid feature detection!
> Nb keypoints detected: 595 for PyramidKAZE method.
> Nb keypoints detected: 455 for KAZE method.
> You should not use AKAZE with Pyramid feature detection!
> Nb keypoints detected: 396 for PyramidAKAZE method.
> Nb keypoints detected: 338 for AKAZE method.
> You should not use SIFT with Pyramid feature detection!
> 
>         Start 114: testKeyPoint-6

modules/core/src/tools/file/vpIoTools.cpp
```diff
@@ -140,24 +140,112 @@ const std::string &vpIoTools::getBuildInformation()
...
+      temp_path.resize(temp_path.size() - 1);
+    }
+  }
+  else {
+    temp_path = "C:\temp";
+    try {
+      vpIoTools::makeDirectory(temp_path);
+    } catch (...) {
+      throw(vpException(vpException::fatalError, "Cannot set temp path to %s", temp_path.c_str()));
+    }
```
s-trinh, line 220:
```suggestion
```

Can be removed I think?

---

You can merge the PR.

PR #971, opened 2022-01-07
Fix SIFT usage with OpenCV 4.5.4 on macOS

### LINKED ISSUE #970: testKeyPoint-5 and testKeyPoint-7 are failing with OpenCV 4.5.4 on MacOS 11
This issue was reported [here](https://github.com/lagadic/visp/runs/4689517063?check_suite_focus=true) and mentioned in PR #965.

The following snippet reproduces the issue on MacOS BigSur 11.6.1 with OpenCV 4.5.4 installed using brew

```
#include <iostream>
#include <string>

#include <opencv2/core.hpp>
#include <opencv2/highgui.hpp>
#include <opencv2/features2d.hpp>

int main()
{
  std::string filename = "../monkey.png";
  cv::Mat img = cv::imread(filename, cv::IMREAD_GRAYSCALE);

  //cv::Ptr<cv::FeatureDetector> detector = cv::FastFeatureDetector::create();
  cv::Ptr<cv::FeatureDetector> detector = cv::SiftFeatureDetector::create();
  cv::Mat mask = cv::Mat::zeros(img.rows, img.cols, CV_8U);
  mask = cv::Mat::ones(img.rows, img.cols, CV_8U) * 255;

  std::vector<cv::KeyPoint> kp;
  detector->detect(img, kp, mask);

  std::cout << "N keypoints: " << kp.size() << std::endl;
}
```
```
$ ./test-sift 
Illegal instruction: 4
```

It clearly shows that the issue reported by ViSP ci is related to OpenCV.

Does this issue occur also with OpenCV 4.5.5 ?

fspindle:
In `testKeyPoint-5.cpp` and `testKeyPoint-7.cpp` SIFT usage is disabled when using OpenCV 4.5.4

This PR is here to fix #970

[x] Fix tested with OpenCV 4.5.5 where SIFT usage is enabled
[x] Fix tested with OpenCV 4.5.4 where SIFT usage is **disabled**
[x] Fix tested with OpenCV 4.5.3 where SIFT usage is enabled

The fix consists right now to turn SIFT usage off in `testKeyPoint-5.cpp` and `testKeyPoint-7.cpp`
when OpenCV 4.5.4 AND macOS since is seems to work on Ubuntu and Windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,15 @@ if(POLICY CMP0075)
   cmake_policy(SET CMP0075 NEW) # For check_include_file and cmake 3.12.0
 endif()
 
+if(APPLE)
+  # Fix following errors for libpng and libjpeg detection:
+  # - Application built with libpng-1.4.12 but running with 1.6.37,
+  # - Wrong JPEG library version: library is 90, caller expects 80.
+  # The reason is that headers were found in /Library/Frameworks/Mono.framework/Headers
+  # while libraries in /usr/local/lib
+  set(CMAKE_FIND_FRAMEWORK LAST)
+endif()
+
 project(VISP C CXX)
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
--- a/ChangeLog
+++ b/ChangeLog
@@ -35,6 +35,7 @@ ViSP 3.4.1 (under development)
     . [#950] Cannot build tutorial-grabber-multiple-realsense.cpp with Visual Studio 2017
     . [#961] Build error on Ubuntu 21.10 with glibc 2.34
     . [#963] Unable to read a sequence of non consecutive images
+    . [#970] testKeyPoint-5 and testKeyPoint-7 are failing with OpenCV 4.5.4 on MacOS 11
     . [#972] Error running some examples: Application built with libpng-1.4.12
              but running with 1.6.37
     . [#973] Error running some examples: Wrong JPEG library version: library is 90,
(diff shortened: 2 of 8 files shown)


s-trinh:
> TBD: Maybe we should turn off SIFT when OpenCV 4.5.4 AND macOS since is seems to work on Ubuntu and Windows

Did you manage to reproduce the issue locally and not only on the CI?

If the issue appears only on the CI, no need to disable SIFT in my opinion unless the issue has been reported on the OpenCV repo (but a very quick search found nothing related?).

fspindle:
@s-trinh I was able to reproduce the issue on my macOS with OpenCV 4.5.4 either installed using brew either build from source.
I don't have the issue when I use OpenCV 4.5.5 build from source.

The snippet mentioned in #970 based only on OpenCV works on my macOS with 4.5.5. It doesn't work with 4.5.4

s-trinh:
> [ ] TBD: Maybe we should turn off SIFT when OpenCV 4.5.4 AND macOS since is seems to work on Ubuntu and Windows

The options in my opinion:
- do nothing since ultimately the user will need to handle this specific case in his code, also maybe the descriptors computation part still works?
- in `vpKeyPoint` do not execute `detector->detect(img, kp, mask);` but  instead throw an exception to the user

I would go with 1).

fspindle:
@s-trinh I don't think that modifying `vpKeyPoint` is a good idea. The problem is that for the packaging with homebrew and also ROS2 that supports now macOS all the test should succeed. That's why I don't see any other solution for the moment than disabling SIFT usage in `testKeyPoint-5` and `testKeyPoint-7` when OpenCV 4.5.4 AND macOS.

On my macbook I'm able to reproduce the issue that occurs only with OpenCV 4.5.4. 
Tests are successful with OpenCV 4.5.3 and 4.5.5 build from source

fspindle:
@s-trinh I will merge this PR to be able to continue preparing visp next release

s-trinh:
@fspindle 

When the next version is expected to be released?

fspindle:
@s-trinh I hope in 2 weeks and at least before end of January to let me the time to package for Debian. The goal is to have the new version in Ubuntu 22.04

PR #984, opened 2022-01-13
Introduce ios workflow

fspindle:


.github/workflows/ios.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-10.15, macos-11.0]
+
+    steps:
+    # https://github.com/marketplace/actions/cancel-workflow-action
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.5.0
```
s-trinh, line 21:
```suggestion
      uses: styfle/cancel-workflow-action@0.9.1
```

For latest version.

fspindle: Yes, I just saw your change in the other workflow files

PR #987, opened 2022-01-18
Feature: stb_truetype interface to load font

s-trinh:
To be able to load TrueType font (`.ttf`), default is [Rubik Fonts](https://github.com/googlefonts/Rubik). Now French accents work :)

Examples:
canvas_color

canvas_gray


cmake/VISPGenerateHeaders.cmake
```diff
@@ -53,6 +53,8 @@ endif()
 if(VISP_INSTALL_DIR_OGRE_RESOURCES)
   list(APPEND VISP_HAVE_OGRE_PLUGINS_PATH "${CMAKE_INSTALL_PREFIX}/${VISP_LIB_INSTALL_PATH}/visp/data/ogre-simulator")
 endif()
+# Rubik-Regular.ttf font for vpFont
+set(VISP_RUBIK_FONT_RESOURCES_PATH "${VISP_SOURCE_DIR}/modules/core/src/image/private/Rubik-Regular.ttf")
```
s-trinh, line 57: This needs to be updated to copy the font to the `build` directory.

Same when using `make install`.

fspindle: @s-trinh Done thanks

PR #996, opened 2022-01-19
Update vpFont TrueType support for old compiler

s-trinh:
Hopefully replace #994

@fspindle 
If you have the time, can you try this PR for compatibility with old platforms on CDash?

### DIFF
--- a/modules/core/src/image/vpFont.cpp
+++ b/modules/core/src/image/vpFont.cpp
@@ -60,9 +60,9 @@
 #include <visp3/core/vpFont.h>
 #include <visp3/core/vpIoException.h>
 #include <visp3/core/vpIoTools.h>
+#include <visp3/core/vpMath.h>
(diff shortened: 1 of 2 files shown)


fspindle:
@s-trinh I merged and then revert PR #994 since I didn't see this PR. I will check tomorrow

modules/core/src/image/vpFont.cpp
```diff
@@ -2172,6 +2172,7 @@ class vpFont::Impl
 
   void LoadTTF(const std::string & filename)
   {
+#if 0 // disable for compatibility with old platform (Ubuntu 12.04)
     std::ifstream fontFile(filename.c_str(), std::ios::binary |std::ios::ate);
```
fspindle, line 2176: I suggest to replace `#if 0` by `#if (VISP_CXX_STANDARD > VISP_CXX_STANDARD_98)`

PR #947, opened 2021-11-21
Update simdlib to 4.9.107 + image IO backend

s-trinh:
TODO:
- [ ] add image encode/decode
- [ ] test image encode with RealSense?
- [ ] use `SimdFree()`
- [ ] fix Java binding
- [ ] fix build
- [ ] check Simd CMakeLists.txt for VS
- [ ] clean code
- [ ] docs

Some benchmark numbers for Simd image loading/writing: https://github.com/ermig1979/Simd/issues/183
The idea is to have a backend option to select which library is used for image loading/writing, with a default backend.
Always build ViSP with stb and Simd support.

ViSP images: https://github.com/lagadic/visp-images/pull/17

PR #1004, opened 2022-01-24
Feature: image I/O backend for JPEG and PNG types

s-trinh:
see also #947

---

Some performance numbers on `Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz CPU` and Ubuntu 20.04:

- `libjpeg-turbo` is actually used, I would also recommend to install `libjpeg-turbo` instead of classical `libjpeg`
- current `Simd` code contains only optimized code for image writing, see: https://github.com/ermig1979/Simd/issues/183#issuecomment-972928613
- `libpng` seems to not perform very well, but we are not comparing file size

Figure_1
Figure_2
Figure_3
Figure_4
Figure_5
Figure_6
Figure_7
Figure_8


modules/io/include/visp3/io/vpImageIo.h
```diff
@@ -138,11 +147,11 @@ class VISP_EXPORT vpImageIo
...
 
-  static void readJPEG(vpImage<unsigned char> &I, const std::string &filename);
-  static void readJPEG(vpImage<vpRGBa> &I, const std::string &filename);
+  static void readJPEG(vpImage<unsigned char> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND);
+  static void readJPEG(vpImage<vpRGBa> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND);
 
-  static void readPNG(vpImage<unsigned char> &I, const std::string &filename);
-  static void readPNG(vpImage<vpRGBa> &I, const std::string &filename);
+  static void readPNG(vpImage<unsigned char> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND);
+  static void readPNG(vpImage<vpRGBa> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND);
```
s-trinh, line 154: Use `int` type for `backend` otherwise Java binding will not work and I don't know how to solve it.

modules/io/include/visp3/io/vpImageIo.h
```diff
@@ -153,10 +162,10 @@ class VISP_EXPORT vpImageIo
...
 
-  static void writeJPEG(const vpImage<unsigned char> &I, const std::string &filename);
-  static void writeJPEG(const vpImage<vpRGBa> &I, const std::string &filename);
+  static void writeJPEG(const vpImage<unsigned char> &I, const std::string &filename, int quality=90, int backend=IO_DEFAULT_BACKEND);
+  static void writeJPEG(const vpImage<vpRGBa> &I, const std::string &filename, int quality=90, int backend=IO_DEFAULT_BACKEND);
 
-  static void writePNG(const vpImage<unsigned char> &I, const std::string &filename);
-  static void writePNG(const vpImage<vpRGBa> &I, const std::string &filename);
+  static void writePNG(const vpImage<unsigned char> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND);
+  static void writePNG(const vpImage<vpRGBa> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND);
```
s-trinh, line 169: Currently only with `JPEG` we can select the quality, but theoretically we should do the same with PNG.

Better, something like a `vector` of options similar to OpenCV `ImwriteFlags`: https://docs.opencv.org/4.5.5/d8/d6a/group__imgcodecs__flags.html#ga292d81be8d76901bff7988d18d2b42ac

```cpp

```

modules/io/test/perfImageLoadSave.cpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+  }
+}
+
+std::string username, directory_filename_tmp;
+
+#if !defined(_WIN32) && (defined(__unix__) || defined(__unix) || (defined(__APPLE__) && defined(__MACH__))) // UNIX
+std::string tmp_dir = "/tmp/";
+#else
+std::string tmp_dir = "C:/Temp/";
```
s-trinh, line 140: Could be updated since now there is `makeTempDirectory()` implemented for Windows.

s-trinh (commented):
Some self-review comments.

modules/io/include/visp3/io/vpImageIo.h
fspindle: ok

modules/io/include/visp3/io/vpImageIo.h
fspindle: - PNG quality could be introduced in a future release.
- For `writeJPEG()`, I suggest to add first the backend, and then the quality

```
  static void writeJPEG(const vpImage<unsigned char> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND, int quality=90);
  static void writeJPEG(const vpImage<vpRGBa> &I, const std::string &filename, int backend=IO_DEFAULT_BACKEND, int quality=90);
```
so that if we add quality to `writePNG()` the prototype remain similar

s-trinh: :+1:

modules/io/src/image/vpImageIo.cpp
```diff
@@ -558,1681 +349,250 @@ void vpImageIo::write(const vpImage<vpRGBa> &I, const std::string &filename)
...
-  Write the content of the image bitmap in the file which name is given by \e
-  filename. This function is built like portable gray pixmap (eg PGM P5) file.
-  but considers float image data.
-
-  \param I : Image to save as a (PFM P8) file.
-  \param filename : Name of the file containing the image.
-*/
-
-void vpImageIo::writePFM(const vpImage<float> &I, const std::string &filename)
+void vpImageIo::readJPEG(vpImage<unsigned char> &I, const std::string &filename, int backend)
```
s-trinh, line 361: Should be updated to not throw an exception if the backend is not available but instead print an error and use either `stb` or `simd` as they are always available depending on the performance numbers.

Same for PNG.

modules/io/test/perfImageLoadSave.cpp
fspindle: Resolved in commit 7c81377

modules/io/include/visp3/io/vpImageIo.h
fspindle: Resolved in commit 9180a98

fspindle:
@s-trinh Thanks. I will merge this PR even if there remain some potential issues.

There is the `testColorConversion.cpp` test that seems to segfault sometimes (not always from what I understand) on `mingw` workflow. I suspect something wrong in `BGRToGrey()` when image size is 64x4. We can come back later to fix it if confirmed.

The sanitizers show also some issues when `-fsanitize=thread` is used.

PR #1018, opened 2022-02-04
Munkres Assignment Algorithm

JuInria:
Munkres Assignment Algorithm (**C++ >= 17**):
  - Class: vpMunkres (`visp/modules/core/include/visp3/core/munkres/vpMunkres.h`)
  - Tests (classic + catch2): `visp/modules/core/test/munkres/testMunkres.cpp`
  - Tutorial: `visp/tutorial/munkres`
  - Doxygen: vpMunkres class + tutorial

PR #1019, opened 2022-02-07
Bump default cxx standard to 17

fspindle:
- We need to allow using modern C++ (see PR #1018)
- Modify workflow to test cxx standard
- Introduce new workflow to test `visp_contrib`, `camera_localization` and `visp_started`


fspindle:
Merged accidentally. Hope it will pass ci

PR #1038, opened 2022-03-14
Add vpRotationMatrix::orthogonalize() to have a proper rotation matrix

s-trinh:
related https://github.com/lagadic/visp/pull/903

---

Without rotation matrix orthogonalization, initializing a `vpRotationMatrix` with custom values must pass [`isARotationMatrix()`](https://visp-doc.inria.fr/doxygen/visp-daily/classvpRotationMatrix.html#a3d3a25cdcfe64aa469bbc2d14502597b) test.

Since often I copy-paste rotation values from other programs, the default print precision can be lower than the required precision in `isARotationMatrix()`.

### DIFF
--- a/modules/core/src/math/transformation/vpRotationMatrix.cpp
+++ b/modules/core/src/math/transformation/vpRotationMatrix.cpp
@@ -398,6 +398,9 @@ vpRotationMatrix &vpRotationMatrix::operator*=(double x)
 
 /*!
   Test if the rotation matrix is really a rotation matrix.
+
+  \return true if the matrix is a rotation matrix, false otherwise.
+
 */
 bool vpRotationMatrix::isARotationMatrix(double threshold) const
 {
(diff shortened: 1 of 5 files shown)


modules/core/src/math/transformation/vpRotationMatrix.cpp
```diff
@@ -924,6 +932,32 @@ vpRotationMatrix vpRotationMatrix::mean(const std::vector<vpRotationMatrix> &vec
   return R;
 }
 
+/*!
+  Perform rotation matrix orthogonalization.
+ */
+void vpRotationMatrix::orthogonalize()
```
s-trinh, line 938: @chaumett

Feel free to review and comment if this approach is correct or not.

chaumett: Very nice! 
From French, I would have used the term "orthonormalize" instead of "orthogonalize" but this is indeed the correct term in english...

fspindle:
@s-trinh Thanks

PR #1043, opened 2022-03-24
Compute object pose from planar object and rgbd camera

JuInria:
## TODO

- [vpPosePlanar](visp/modules/vision/src/pose-estimation/vpPosePlanar.cpp) is similar to [vpPoseRGBD](visp/modules/vision/src/pose-estimation/vpPoseRGBD.cpp) but more generic (it needs also c++17): check redundancy | vpPoseRGBD should be replaced by vpPosePlanar + vpPlaneEstimation
- Use vpException instead of optional to reduce CXX minimal version:
  -  [vpPlaneEstimation](visp/modules/vision/src/plane-estimation/vpPlaneEstimation.cpp) can be c++11 compliant (warn: structured bindings will have to be removed or conditioned)
  - [vpPlaneEstimation](visp/modules/vision/src/plane-estimation/vpPlaneEstimation.cpp) can be OpenMP 2.0 compatible (see [here](https://microeducate.tech/c-openmp-parallel-for-loop-alternatives-to-stdvector-closed/))
  -  [vpPosePlanar](visp/modules/vision/src/pose-estimation/vpPosePlanar.cpp) can be c++98 or c++11 compliant (warn: structured bindings will have to be removed or conditioned)
- Test  vpColorDepthConversion::projectColorToDepth
- Import 2pts pose estimation from react


modules/core/src/camera/vpColorDepthConversion.cpp
```diff
@@ -0,0 +1,255 @@
...
+ * \param[in] width : Image width.
+ * \param[in] height : Image height.
+ * \return Adjusted 2D point (<=> [i,j] in [0,height][0,width] ).
+ */
+vpImagePoint adjust2DPointToBoundary(const vpImagePoint &ip, double width, double height)
+{
+#if (VISP_CXX_STANDARD > VISP_CXX_STANDARD_98)
+  return {vpMath::clamp(ip.get_i(), 0., height), vpMath::clamp(ip.get_j(), 0., width)};
+#else
+  vpImagePoint(vpMath::clamp(ip.get_i(), 0., height), vpMath::clamp(ip.get_j(), 0., width));
```
JuInria, line 69: **return** vpImagePoint(vpMath::clamp(ip.get_i(), 0., height), vpMath::clamp(ip.get_j(), 0., width));

modules/core/test/tools/geometry/testPolygon2.cpp
```diff
@@ -0,0 +1,107 @@
...
+  {
+    const vpRect rect{0, 0, 200, 400};
+    const std::vector<vpImagePoint> rect_corners{rect.getTopLeft(), rect.getTopRight(), rect.getBottomRight(),
+                                                 rect.getBottomLeft()};
+
+    vpPolygon poly{};
+    poly.buildFrom(rect_corners, true);
+
+    for (const auto &poly_corner : poly.getCorners()) {
+      REQUIRE(std::find(cbegin(rect_corners), cend(rect_corners), poly_corner) != cend(rect_corners));
```
JuInria, line 63: Should be something like:

```
#if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_14)
REQUIRE(std::find(cbegin(rect_corners), cend(rect_corners), poly_corner) != cend(rect_corners));
#else if(VISP_CXX_STANDARD >= VISP_CXX_STANDARD_11)
REQUIRE(std::find(begin(rect_corners), end(rect_corners), poly_corner) != end(rect_corners));
#else
REQUIRE(std::find(rect_corners.begin(), rect_corners.end(), poly_corner) != rect_corners.end());
#endif
```

modules/core/src/camera/vpColorDepthConversion.cpp
fspindle: Yes

modules/core/test/tools/geometry/testPolygon2.cpp
JuInria: Furthermore `for (const auto &poly_corner : poly.getCorners())` should be replaced by `for (std::vector<vpImagePoint>::const_iterator it = poly.getCorners().begin(); it != poly.getCorners().end(); ++it)` for c++98 support.

modules/core/include/visp3/core/vpMath.h
```diff
@@ -121,6 +122,25 @@ class VISP_EXPORT vpMath
...
+    Clamp a value to boundaries.
+    \param v : The value to clamp.
+    \param lower, upper : The boundaries to clamp `v` to.
+
+    Throw a vpException if the value of `lower` is greater than `upper`.
+  */
+  template <typename T> static inline T clamp(const T &v, const T &lower, const T &upper)
+  {
+#if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_17)
+    return std::clamp(v, lower, upper);
```
JuInria, line 135: `#include <algorithm>` could be missing for `std::clamp`

fspindle: Already included

PR #1063, opened 2022-05-30
Update calibrate-camera app with calibration board occupancy in color

s-trinh:
Update code to display the calibration board occupancy with a colormap and with squares for the extracted calibration points:

image

Display side-by-side the fitted lines for the distorted and the undistorted images:

image


fspindle:
@s-trinh Thanks

PR #1065, opened 2022-06-05
Add figures and formulas for NED and ENU coordinates systems

s-trinh:
Figure from: https://en.wikipedia.org/wiki/File:ECEF_ENU_Longitude_Latitude_relationships.svg

See also: https://www.newbooks-services.de/MediaFiles/Texts/4/9780857296344_Excerpt_002.pdf


fspindle:
@s-trinh Thanks, ok for me

s-trinh:
👍 
Thanks for the review.

PR #1086, opened 2022-08-06
Add vpThetaU rotation multiplication

s-trinh:
To be able to directly chain rotations using the theta.u rotation representation.
It avoids converting to another rotation representation.


fspindle:
@s-trinh Thanks

PR #1091, opened 2022-09-12
Add tests to check image I/O using the different backends

s-trinh:
I have some segfault on my computer.
Let's see if it can be reproduced on the CI.

---

TODO:
- ~~[ ] Update Simdlib to a more recent version?~~
- [x] Add better tests for image I/O?
- [x] Print in the CMake config output some info about the CPU (useful to be able to debug Simd lib)? _But not sure if it is working on Windows?_
- [x] Print in the CMake config output the version of the embedded Simd lib?


fspindle:
- We are using Simd version `4.9.109`, latest version is `5.1.117`
```
Use Simd:                    yes (ver 4.9.109)
```
- Yes we should upgrade to the lastest version, but if I remmember well this operation is always tricky...
- In the mean time I will add new printings to catch CPU info

s-trinh:
Some warnings:

```bash
 /home/runner/work/visp/visp/modules/tracker/tt_mi/src/tools/vpTemplateTrackerMIBSpline.cpp: In static member function 'static void vpTemplateTrackerMIBSpline::PutTotPVBspline4(double*, double*, double*, int, double, int, double, int, double*, unsigned int&)':
/home/runner/work/visp/visp/modules/tracker/tt_mi/src/tools/vpTemplateTrackerMIBSpline.cpp:432:29: warning: '#'target_mem_ref' not supported by dump_expr#<expression error>' is used uninitialized in this function [-Wuninitialized]
  432 |       Prt[ind + it] += Br * *ptBti;
      |                             ^~~~~~
/home/runner/work/visp/visp/modules/tracker/tt_mi/src/tools/vpTemplateTrackerMIBSpline.cpp:435:33: warning: '#'target_mem_ref' not supported by dump_expr#<expression error>' is used uninitialized in this function [-Wuninitialized]
  435 |       double Br_ptdBti_ = Br * (*ptdBti);
      |                                ~^~~~~~~~
/home/runner/work/visp/visp/modules/tracker/tt_mi/src/tools/vpTemplateTrackerMIBSpline.cpp:436:34: warning: '#'target_mem_ref' not supported by dump_expr#<expression error>' is used uninitialized in this function [-Wuninitialized]
  436 |       double Br_ptd2Bti_ = Br * (*ptd2Bti);
      |                                 ~^~~~~~~~~
```

---

The newer version of [Simd](https://github.com/ermig1979/Simd/releases) removed the support of `SSE2` (replaced by `SSE4.1`).
What is the CPU model on the machine with Windows XP in the lab?

CMakeLists.txt
```diff
@@ -1603,6 +1604,92 @@ status("")
...
+status("    Number of CPU physical cores:"                 "${NUMBER_OF_PHYSICAL_CORES}")
+status("    Total physical memory (in MiB):"               "${TOTAL_PHYSICAL_MEMORY}")
+status("    OS name:"                                      "${OS_NAME}")
+status("    OS release:"                                   "${OS_RELEASE}")
+status("    OS version:"                                   "${OS_VERSION}")
+status("    OS platform:"                                  "${OS_PLATFORM}")
+status("    CPU name:"                                     "${PROCESSOR_NAME}")
+status("    Is the CPU 64-bit?"                            ${IS_64BIT} THEN "yes" ELSE "no")
+status("    Does the CPU have FPU?"                        ${HAS_FPU} THEN "yes" ELSE "no")
+status("    SSE2?"                                         ${HAVE_SSE2_FLAG} THEN "yes" ELSE "no")
```
s-trinh, line 1684: Maybe it should rather be?

```suggestion
status("  Compiler flags support:")
status("    SSE2 flag?"                                         ${HAVE_SSE2_FLAG} THEN "yes" ELSE "no")
status("    SSE3 flag?"                                         ${HAVE_SSE3_FLAG} THEN "yes" ELSE "no")
status("    SSSE3 flag?"                                        ${HAVE_SSSE3_FLAG} THEN "yes" ELSE "no")
status("    SSE4.1 flag?"                                       ${HAVE_SSE4_1_FLAG} THEN "yes" ELSE "no")
status("    SSE4.2 flag?"                                       ${HAVE_SSE4_2_FLAG} THEN "yes" ELSE "no")
status("    AVX flag?"                                          ${HAVE_AVX_FLAG} THEN "yes" ELSE "no")
status("    AVX2 flag?"                                         ${HAVE_AVX2_FLAG} THEN "yes" ELSE "no")
status("    NEON flag?"                                         ${HAVE_NEON_FLAG} THEN "yes" ELSE "no")
```

If `vp_check_compiler_flag` only check for correct compilation result and does not execute the code, it is indeed `Compiler flags support`.

fspindle:
@s-trinh I changed the code to check cpu optimisation like `SSE2`, `SSE3`,...`NEON` rather than checking the flags.

It gives on my Mac the following:

```
```

fspindle:
On win10 used in the lab for ci it produces output like:
```
  System information:
    Number of CPU logical cores: 4
    Number of CPU physical cores: 4
    Total physical memory (in MiB): 16057
    OS name:                     Windows
    OS release:                  Professional
    OS version:                  (Build 19044)
    OS platform:                 AMD64
    CPU name:                    Pentium II (0.25 micron)
    Is the CPU 64-bit?           yes
    Does the CPU have FPU?       yes
    CPU optimization:            SSE2 SSE3 SSSE3 SSE4_1 SSE4_2
```

fspindle:
@s-trinh This is the Windows XP machine configuration

```
  Platform:
    Timestamp:                   2022-09-27T16:32:17Z
    Host:                        Windows 5.1.2600 x86
    CMake:                       3.11.4
    CMake generator:             Visual Studio 10 2010
    CMake build tool:            C:/WINDOWS/Microsoft.NET/Framework/v4.0.30319/MSBuild.exe
    MSVC:                        1600

  System information:
    Number of CPU logical cores: 2
    Number of CPU physical cores: 2
    Total physical memory (in MiB): 2031
    OS name:                     Windows
    OS release:                  XP Professional
    OS version:                  Service Pack 3 (Build 2600)
    OS platform:                 x86
    CPU name:                    Unknown P6 family
    Is the CPU 64-bit?           no
    Does the CPU have FPU?       yes
    CPU optimization:            SSE2 SSE3 SSSE3 SSE4_1 SSE4_2
```

s-trinh:
@fspindle 
Many thanks for the updates.

---

The newer version of Simd removed the support of SSE2 instruction sets (https://github.com/ermig1979/Simd/releases).
Minimal instruction sets are now SSE4.1. Since the Windows XP machine supports SSE4.1 instruction sets, it means that future upgrade of the Simd lib will not penalize this machine (Simd lib always providing baseline implementation (no vectorization)).

---

Sorry just to be sure, could you copy the following information for the CPU model of the `Windows XP` machine:
- `win + R` then `msinfo32.exe` and OK
- tutorial: https://answers.microsoft.com/en-us/windows/forum/all/where-do-i-find-system-information-for-windows-xp/32f85fc5-2866-45b4-ac15-58a3a22c9f47

Because if I have understood correctly the `vp_check_compiler_flag()` macro, it does not execute the program, only check the correct build of the code?

Most likely the CPU on this machine supports `SSE4.1/SSE4.2` instruction sets, but I want to be sure.

---

BTW, I am using [Meld](https://meldmerge.org/) as a visual diff tool. Comparing large diff when upgrading to a newer version of Simd is not so difficult I would say.

fspindle:
@s-trinh I have a simdlib PR in preparation to introduce some changes that we did for ViSP. It should be ready end of the week. If merged, it could ease simdlib upgrade

s-trinh:
You can merge this PR.

---

But for curiosity, one day if you could copy / paste the model of the CPU of the Windows XP.

fspindle:
@s-trinh Below is the output of `msinfo32`

| Element | Value |
| ---- | ---- |
|OS Name |Microsoft Windows XP Professional|
|Version| 5.1.2600 Service Pack 3 Build 2600|
|OS Manufacturer| Microsoft Corporation|
|System Model| RX965QV|
|System Type| X86-based PC|
|Processor| x86 Family 6 Model 15 Stepping 13 GenuineIntel ~2194 Mhz|
```

fspindle:
@s-trinh Following our discussion I opened the [simd PR](https://github.com/ermig1979/Simd/pull/232)

s-trinh:
> System Model 	RX965QV
> Processor 	x86 Family 6 Model 15 Stepping 13 GenuineIntel ~2194 Mhz

From [this site](https://en.wikichip.org/wiki/intel/cpuid), it gives:

image

And thanks to [this site](https://nikkhokkho.sourceforge.io/static.php?page=Lamark-Results), it gives:

> Intel(R) Core(TM)2 Duo CPU E4500 @ 2.20GHz

Comfirmed with the [motherboard model](https://www.bcmcom.com/admin/Installation/RX965QV_QIG_1.03.pdf).

The specs of this processor gives:

> MMX, SSE, SSE2, SSE3, SSSE3, EM64T

From these sites:
- https://valid.x86.fr/e5yv8l
- https://www.cpu-world.com/CPUs/Core_2/Intel-Core%202%20Duo%20E4500%20HH80557PG0492M%20(BX80557E4500).html

---

Newer version of the Simd lib removed the support of `SSE2`. So we would need to pay attention to that and keep the `SSE2` files to not penalize this computer in theory.

fspindle:
The win XP computer was not used since more than 2 years, and I don't think that it will be used in the future.
Just in case of, I think that ensuring that ViSP builds on that platform is a good idea, but we can than make the assumption that using `SimdBase` is good enough. For me there is no need to keep the support of `SSE2` and thus keep the `SSE2` files.

s-trinh:
Understood.
Since newer versions of Simd use SSE4.1 as minimum, it should give a little better performance (not as much as using AVX instructions set, which uses 256-bits wide instead of 128-bits wide).

PR #1097, opened 2022-09-28
Update vpFeatureThetaU.cpp and PBVS examples

jifu23:
According to the paper, Lw should be -[theta/2 u]_x.

In PBVS examples, convergence_threshold_tu is in rad, but error_tu was in degree, which makes convergence very difficult.

### DIFF
--- a/example/servo-franka/servoFrankaPBVS.cpp
+++ b/example/servo-franka/servoFrankaPBVS.cpp
@@ -335,7 +335,7 @@ int main(int argc, char **argv)
         vpTranslationVector cd_t_c = cdMc.getTranslationVector();
         vpThetaUVector cd_tu_c = cdMc.getThetaUVector();
         double error_tr = sqrt(cd_t_c.sumSquare());
-        double error_tu = vpMath::deg(sqrt(cd_tu_c.sumSquare()));
+        double error_tu = sqrt(cd_tu_c.sumSquare());
 
         ss.str("");
         ss << "error_t: " << error_tr;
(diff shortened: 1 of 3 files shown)


s-trinh:
This is just an idea.
I was wondering if an updated version of the tutorial paper could be uploaded on open access sites like?
- https://hal.archives-ouvertes.fr/inria-00350283
- https://hal.inria.fr/inria-00350283
- http://rainbow-doc.irisa.fr/pdf/2006_ieee_ram_chaumette.pdf

Or an erratum published somewhere?

jifu23:
> Hi @jifu23
> 
> Thanks, but I cannot merge your PR.
> 
> The equation (14) you mention in this [paper](http://rainbow-doc.irisa.fr/pdf/2006_ieee_ram_chaumette.pdf) is wrong.
> 
> The equations given in [ViSP doc](https://visp-doc.inria.fr/doxygen/visp-daily/classvpFeatureThetaU.html) are the right ones.
> 
> Concerning `error_tu`, this vector is converted in degrees to display human understandable feedback.
> 
> The error is more in `convergence_threshold_tu` that should be initialized with a value in degree. I did the [changes](https://github.com/lagadic/visp/pull/1098/commits/5f01e61aa152186343ddd9e26ce3726fc0cb3cc8).

Thanks for your reply, that‘s a better way indeed.

fspindle:
An updated version of the paper is available here: http://rainbow-doc.irisa.fr/pdf/2006_ieee_ram_chaumette.pdf

Updating other download locations will take more time.

PR #1103, opened 2022-09-30
Introduce Ubuntu 22.04 and macOS 12.0 in github action ci

fspindle:


fspindle:
For now ViSP doesn't support iOS framework on macOS 12.0 with Xcode 14 (see #1107)
The same issue occurs with OpenCV. An issue is already opened there.

AndreV84:
@fspindle Salut! Ca Va?
est ce possible : visp ios/android unity wrapper apriltag implementation?
Merci Bien

PR #1122, opened 2023-01-10
Improve MAVSDK support   

Tonirayn:
## Fixed take off: 
It presented a blocking behaviour activating only once the offboard mode. If the mavsdk don't receive a fresh input in a few ms it deactivates the offboard mode and the visp side will block trying to halt the offboard mode. 

## Fixed Odometry get:
Before the method was using the raw odometry data and that, with SITL was causing issues since it is only available GPS.   

## Added get home position:
Added a method to retrieve the home position. Used with SITL in a multi-drone context, it allows reading the pose in the correct location.


fspindle:
The last changes bring the following capabilities:
- compat with odometry coming from GPS or MoCAP
- new function to enable/disable off-board or guided with `takeControl()` and `releaseControl()` functions respectively
- new behavior for goto functionality with `setPosition()` and `setPositionRelative()` functions that have a flag indicating if the function is blocking or not. By blocking we meant that the function waits until the vehicle reaches the desired position.
- timeout introduced in `takeoff()` and positioning functions
- compat with rover equipped with Pixhawk
- various fixes

PR #1123, opened 2023-01-23
Fix LonLatCamPosesVisualizer.py script

s-trinh:
Some fixes when running the LonLatCamPosesVisualizer.py script on recent version of Ubuntu / Python3.


fspindle:
Thanks

PR #1128, opened 2023-02-16
Support of EXR image file format

s-trinh:
Suggestion to support the EXR image file format using the [Tiny OpenEXR image library](https://github.com/syoyo/tinyexr).
From my point of view, advantages would be:
- standard file format to support I/O of floating-point data such as depth maps
- can provide zlib data compression
- should replace the ViSP `.bin` format used for depth map support, and in the same time should avoid issue with little-endian / big-endian


fspindle:
@s-trinh Nice thanks. Can you introduce the following images in `visp-images/memorial`
```
memorial_color_32bits.exr
memorial_color_16bits.exr
memorial_gray_32bits.exr
memorial_gray_16bits.exr
```

s-trinh:
Somehow using FreeEXRImage() and FreeEXRHeader() segfault the program?
Let see if the builders report some memory leak.

---

Looks like free() should be used: https://github.com/syoyo/tinyexr/blob/02310c77e5156c36fedf6cf810c4071e3f83906f/test/unit/tester.cc#L1311-L1313 ?

fspindle:
@s-trinh I suggest to merge this branch with lagadic/visp master branch.

s-trinh:
@fspindle 

If all the checks are successful and if the code is correct, the PR can be merged along with https://github.com/lagadic/visp-images/pull/23.

For now, big endian is not totally supported, see: https://github.com/syoyo/tinyexr/issues/189.

PR #1133, opened 2023-03-03
Added optional parameters for drawFrame allowing to display along wit…

rolalaro:
Modification to vp_display_display_frame : 
- Changed its implementation due to visible errors when the camera has no distortion
- Added optional parameters to this method and vpDisplay::displayFrame in order to write the name of the frame beside the frame. The direction of the offset between the name and the frame is computed to avoid as much as possible crossing an axis.
- A tutorial tutorial-draw-frame has been added in tutorials/draw to show an example of how the method behaves.


fspindle:
@rolalaro Thanks

PR #1137, opened 2023-03-16
"Fix" poseDementhon* hard-coded pseudo-inverse threshold

rolalaro:
[CORPS] Added a dementhonSvThresh attribute, a setter and using it when computing pseudo-inverse in poseDementhon*
It can be useful when your application require a fine-tuning of the pseudo-inverse svd threshold to determine if the configuration is planar or not

### DIFF
--- a/modules/vision/src/pose-estimation/vpPoseDementhon.cpp
+++ b/modules/vision/src/pose-estimation/vpPoseDementhon.cpp
@@ -493,7 +493,7 @@ void vpPose::poseDementhonPlan(vpHomogeneousMatrix &cMo)
     errorMsg << "In Dementhon planar, after ";
     errorMsg << nbMaxIter;
     errorMsg << " trials multiplying the svd threshold by ";
-    errorMsg << nbMaxIter;
+    errorMsg << svdFactorUsedWhenFailure;
     errorMsg << ", rank (";
     errorMsg << irank;
     errorMsg << ") is still not 3";
--- a/modules/vision/src/pose-estimation/vpPoseDementhon.cpp
+++ b/modules/vision/src/pose-estimation/vpPoseDementhon.cpp
@@ -469,11 +469,11 @@ void vpPose::poseDementhonPlan(vpHomogeneousMatrix &cMo)
   }
   vpColVector sv;
   vpMatrix Ap, imA, imAt, kAt;
-  bool isRankEqualTo3(false); // Indicates if the rank of A is the expected one
+  bool isRankEqualTo3 = false; // Indicates if the rank of A is the expected one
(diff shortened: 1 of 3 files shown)


fspindle:
@rolalaro Sounds a good idea. Thanks

modules/vision/src/pose-estimation/vpPoseDementhon.cpp
```diff
@@ -466,9 +469,34 @@ void vpPose::poseDementhonPlan(vpHomogeneousMatrix &cMo)
   }
   vpColVector sv;
   vpMatrix Ap, imA, imAt, kAt;
-  int irank = A.pseudoInverse(Ap, sv, 1.e-6, imA, imAt, kAt);
-  if (irank != 3) {
-    throw(vpException(vpException::fatalError, "In Dementhon planar, rank is not 3"));
+  bool isRankEqualTo3(false); // Indicates if the rank of A is the expected one
+  double logNofSvdThresh = std::log(dementhonSvThresh)/lnOfSvdFactorUsed; // Get the log_n(dementhonSvThresh), where n is the factor by which we will multiply it if the svd decomposition fails.
```
fspindle, line 473: Maybe CI will break when using c++98: https://cdash-ci.inria.fr/index.php?project=ViSP
If this is the case, we will have to replace with `bool isRankEqualTo3 = false;` and below `int irank = 0;`

modules/vision/src/pose-estimation/vpPoseDementhon.cpp
```diff
@@ -466,9 +469,34 @@ void vpPose::poseDementhonPlan(vpHomogeneousMatrix &cMo)
...
+  
+  if (!isRankEqualTo3) {
+    std::string errorMsg("In Dementhon planar, after ");
+    errorMsg += std::to_string(nbMaxIter);
+    errorMsg += std::string(" trials multiplying the svd threshold by ");
+    errorMsg += std::to_string(nbMaxIter);
+    errorMsg += std::string(", rank (");
+    errorMsg += std::to_string(irank);
+    errorMsg += std::string(") is still not 3");
+    throw(vpException(vpException::fatalError, errorMsg));
```
fspindle, line 499: Same here. `std::to_string` is c++11.
With c++98 we should use `std::stringstream`.

In the next release we will keep c++98 compat. It will be the last release compatible with  c++98.

modules/vision/src/pose-estimation/vpPoseDementhon.cpp
rolalaro: Sorry, did not think of being c++98 compatible.
Changed in commit 279c1e7acb7e675682c481644b95263774853f0a , lines 472 and 476

modules/vision/src/pose-estimation/vpPoseDementhon.cpp
rolalaro: Again, did not think about being c++98 compatible.
Changed in 279c1e7acb7e675682c481644b95263774853f0a , lines 492 to 500

PR #1147, opened 2023-03-27
Update appveyor to use OpenCV 4.7.0

fspindle:


fspindle:
PR closed since ci enters in a timeout after 1 hour.
Solution would we to introduce Visual Studio 2019 and 2022 with last OpenCV release in github workflow

PR #1159, opened 2023-04-17
Update readme file

s-trinh:
- Update Sanitizer url
- Always use https urls (http url is actually automatically redirected to https url)
- Update Q&A forum url

---

image

I think there are some interesting questions. Unfortunately, the questions have not been saved by the Wayback Machine.
If there is a back-up by the Inria, maybe some questions could be transferred to the https://github.com/lagadic/visp/discussions url?


fspindle:
@s-trinh Thanks. Unfortunately questions/answer from old forum web site are lost.

PR #1189, opened 2023-07-05
Fix ci visp_sample build using visp-config script

fspindle:


fspindle:
This PR was needed since changes introduced in [visp_sample/Makefile.visp-config](https://github.com/lagadic/visp_sample/blob/master/Makefile.visp-config)

PR #1192, opened 2023-07-10
Fix invalid usage of OGRE_DIR in CMakeLists.txt

olivier-roussel:
The `OGRE_DIR` cmake variable seems not correctly used here, as it will be set to the directory containing the configuration cmake file (as defined here: https://cmake.org/cmake/help/latest/command/find_package.html#full-signature ).
In other words, the `OGRE_DIR` variable typically already ends up with a `/cmake` directory, so the `CMAKE_MODULE_PATH` should not be appended by an extra `/cmake` directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,7 +335,7 @@ endif(WIN32)
 
 if(UNIX)
   list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/cmake")
-  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/Cmake")
+  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/CMake")
   list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
   list(APPEND CMAKE_MODULE_PATH "/usr/local/lib/OGRE/cmake")
   list(APPEND CMAKE_MODULE_PATH "/usr/lib/OGRE/cmake")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -329,12 +329,13 @@ endif()
 
 # add the path to detect Ogre3D
 if(WIN32)
-  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
+  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/cmake")
   list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
 endif(WIN32)
 
 if(UNIX)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -329,13 +329,13 @@ endif()
 
 # add the path to detect Ogre3D
 if(WIN32)
-  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/cmake")
-  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/CMake")
+  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
+  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
```
fspindle, line 333: Hi, thanks for this PR.
I would prefer that you add a new path instead of replacing that could lead to side effects.
Your 2 lines are the same.

Since windows is unable to make difference between lower and upper case, put simply 
```
list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/cmake")
list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
```

CMakeLists.txt
```diff
@@ -329,13 +329,13 @@ endif()
...
 endif(WIN32)
 
 if(UNIX)
-  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/cmake")
-  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/CMake")
+  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
+  list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
   list(APPEND CMAKE_MODULE_PATH "/usr/local/lib/OGRE/cmake")
```
fspindle, line 339: and here:

```
list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/cmake")
list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/CMake")
list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}")
```

CMakeLists.txt
olivier-roussel: My bad, did not paid attention that it lead to duplicated lines.
Ok to me to keep both `OGRE_DIR` paths (with and without the `cmake` suffix).

CMakeLists.txt
fspindle: should be `list(APPEND CMAKE_MODULE_PATH "${OGRE_DIR}/CMake")`

PR #1205, opened 2023-07-17
Feat github action for windows with clang

fspindle:


fspindle:
Done in #1201 that is more complete

PR #1213, opened 2023-08-09
Canny auto threshold

rolalaro:
Added several features:
- a vpImageFilter::median that computes the median of a single- or multi- channel image
- the possibility to ask the automatic computation of the Canny edge detector thresholds
- update of the tutorial accordingly
- update of the tests to test the vpImageFilter::median methods


fspindle:
Ready to merge when ci ok

PR #1215, opened 2023-08-11
Added a ViSP implementation of Canny edge detector

rolalaro:
Implemented a ViSP version of the Canny edge detector that does not rely on OpenCV imgproc


fspindle:
Ready to merge when ci ok

PR #1216, opened 2023-08-11
Megapose improvements

SamFlt:
This PR brings some improvements to Megapose:

- A fix for network communication in vpMegaPose on Windows
- Notebook version made install fail
- Add a verbose mode to print timings, showing render and model times
- The client's disconnection from the server is now cleaner and does not generate an exception on the server
- Added some documentation on using megapose on Windows: the server should be run in WSL, as it seems that there are some issues with panda3D on Windows
- Added some documentation on the PBVS servoing example and referenced it in the megapose tracking tutorial
- The default model is now RGB, as it is the most usable one (in terms of inference speed)
- During install, if the user uses relative paths, they are converted to absolute ones in order for the megapose server to run from any directory


fspindle:
Ready to merge when ci ok

PR #1217, opened 2023-08-21
Feat dataset generation

SamFlt:
This PR introduces a new script, that uses blenderproc to generate synthetic data from 3D models. This tool can be used to train a YoloV7 to detect certain objects. Other modalities, such as poses, depth, normal maps and segmentation can also be recovered.

The accompanying tutorial explains the different tunable parameters, how to run the script and how to use its output to train a YoloV7-tiny to detect the objects. 

Before merging, some result videos should be uploaded to Youtube and added to the documentation to showcase the detection results.


fspindle:
Ready to merge after final check

PR #1227, opened 2023-09-04
Introduce modern way to link with pthread library

fspindle:
These changes are related to Conda packaging where the following issues was encountered when visp conda package is used as a third-party. In `VISPModules.cmake` there remain the full absolute path of the `m` and `pthread` libraries used during the package generation. When installing the conda package on a linux target, conda is unable to replace these locations with the corresponding ones on the target.
The fix consists in:
- for lib `m` to export the library as private (Fixed in PR # 1226 commit e2ed2225ca50652c4)
- for lib `pthread` to use modern cmake (commit 9de4530c6b95110 and d6e5a2fe74defc6ff6)
- for `openmp` that has a dependency to `pthread` to consider the specific case of `pthread` as above (commit 9209898bd4223) 
- as a side effect, `ms2_32.lib` that was part of `VISP_LINKER_LIBS` that contains now only private libraries, was removed from `VISP_LINKER_LIBS` and considered as a classical 3rd party linked to `visp_core` module (see commit 2c074db)

See issue #1228


fspindle:
@olivier-roussel Moving `VISP_LINKER_LIBS` as private introduced a side effect on Windows where `ws2_32.lib` was not exported as 3rd party. Fixed in commit [2c074db](https://github.com/lagadic/visp/pull/1227/commits/2c074dbdba519e439647e29706118c6d65c05bee)

Without this change, `vpNetwork` class is not usable in a project on windows with ViSP as a 3rd party

fspindle:
Closes #1228

PR #1243, opened 2023-09-28
[tutorial/calibration] Add option to make executable non interactive

florent-lamiraux:
in tutorial-chessboard-pose.
  Install programs tutorial-chessboard-pose and tutorial-hand-eye-calibration
  and rename them "tutorial-" -> "compute-".


fspindle:
@florent-lamiraux Thanks for this PR. Tutorials and examples are not meant to install. But you are right, the two tutorials dedicated to calibration could be useful if installed. Based on this PR I will create a specific folder, maybe called `apps` that will collect some of the tools that are today in `tutorial` and `example` folders.

PR #1249, opened 2023-10-09
Fix gradients canny cht

rolalaro:
- Choice between the OpenCV and ViSP backend for the Canny computation
- Computation of the gradient using Gaussian blur + Sobel instead of derivativeGaussianFillter
- New method of Canny auto threshold


rolalaro:
Remaining work:
- [ ]  Add a tutorial for Canny
- [ ] Add a method to compute the partial normalized gradients

PR #1278, opened 2023-11-21
Fix canny

rolalaro:
Improved the ViSP Canny implementation by :
- changing the edge thinning algorithm
- considering only the gradient points having a value greater than the lower threshold

To test with the result of Canny by OpenCV beside the ViSP implementation result:
```
./tutorial-canny -t -1 -1 -r 0.4 0.95 -b opencv-backend --image
```

Here is a dataset that can be used to have a look at the capabilities of the new implementation:
[testCanny.zip](https://github.com/lagadic/visp/files/13424676/testCanny.zip)


fspindle:
@rolalaro Seems ok to me. I have just the feeling that using `-r 0.4 0.85` gives results much closer to the one obtained with OpenCV backend.

Could be seen especially on `cube.ppm` image. On `mire.png` the values are more around `-r 0.4 0.9`.

Anyway, I will merge.

PR #1287, opened 2023-11-30
Fix canny

rolalaro:
Fix two problems : 
- in vpCannyEdgeDetection, fix the way the gradient orientation is computed, getting results that are much closer to the ones obtained using opencv-backend
- in vpImageFilter, fix a mistake on the way the partial derivatives are normalized when using opencv-backend

### DIFF
--- a/modules/core/src/image/vpCannyEdgeDetection.cpp
+++ b/modules/core/src/image/vpCannyEdgeDetection.cpp
@@ -314,13 +314,15 @@ getManhattanGradient(const vpImage<float> &dIx, const vpImage<float> &dIy, const
 }
 
 /**
- * @brief Get the absolute value of the gradient orientation.
+ * @brief Get the gradient orientation, expressed in degrees, between 0 and +180 degrees.
+ * If the gradient orientation is negative, we add 180 degrees (i.e. M_PI radiants) in
+ * order to keep the same orientation but in the positive direction.
  *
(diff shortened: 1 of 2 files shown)


modules/core/src/image/vpCannyEdgeDetection.cpp
```diff
@@ -235,7 +235,7 @@ vpCannyEdgeDetection::performFilteringAndGradientComputation(const vpImage<unsig
 /**
  * \brief Get the interpolation weights and offsets.
  *
- * \param[in] absoluteTheta : The absolute value of the angle of the edge, expressed in degrees.
+ * \param[in] positiveTheta : The positive value of the angle of the edge, expressed in degrees.
```
fspindle, line 238: To be complete, you could precise that this angle is between -180 and +180 deg

fspindle: Is there a goo reason to consider degrees instead of rad?
antan2() is returning rad
You can use `M_PI_2` and `M_PI_4` macros

fspindle: I suggest to rename `positiveTheta` into `gradientOrientation` to be more explicit

rolalaro: These suggestions have been addressed in the new commit e459437c747354287392d4a6761cb5393aaea853

PR #1301, opened 2024-01-08
Disable Python bindings for old cmake versions

SamFlt:
This PR disables Python bindings for cmake versions that are below 3.19. This version is required because we generate a JSON file that is parsed by the bindings generator.

@rolalaro should test this before merging :)


rolalaro:
@SamFlt OK everything works well ! =)

PR #1311, opened 2024-01-17
Use of boolean mask

rolalaro:
- Possible to use a boolean mask along with vpImageFilter methods to compute desired points only
- Possible to use a boolean mask along with vpHistogram
- Possible to use a boolean mask along with vpCannyEdgeDetector
- Possible to use a boolean mask along with vpCircleHoughTransform


modules/imgproc/src/vpCircleHoughTransform.cpp
```diff
@@ -364,12 +374,9 @@ void vpCircleHoughTransform::computeVotingMask(const vpImage<unsigned char> &I,
 #if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_17)
     mask = std::nullopt;
     opt_votingPoints = std::nullopt;
-#elif (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_11)
```
rolalaro: @fspindle I see that you deleted the NULL rows for C++98, did you add the "hack"
to have nullptr in C++98 that you showed me in ViSP ?

PR #1312, opened 2024-01-19
Introduce tutorial to explain how to build python bindings

fspindle:

PR #1316, opened 2024-01-24
[WIP] Python build improvements

SamFlt:
This PR brings several small improvements to the Python wrapper build step

- [x] The _visp target (Pybind target) is no longer built when calling `make`. Thus, to build the python bindings, one should call `make visp_python_bindings`
- [x]  Fixed some warnings about unintialized variables in bindings compilation. This was due to output parameters not being initialized when declaring them in lambda functions. Basic types such as ints or floats are initialized to 0.
- [x] Update RST documentation with some error messages that can appear when building the bindings. This should probably be moved to the .dox documentation
- [x] Detect and forbid system Python: We cannot mix pip installable packages with system-wide packages (risk of failure, forbidden from 3.12).  This was tested in conda and seems to work, but should be tested with virtualenv. The current solution also requires the CMake Cache to be cleared, otherwise it may fail.
- [x] Ignore `vpAROgre`: this would require to fully wrap the OGRE classes, and generates some error at compile time
- [x] Improve name resolution for types, that could result in errors before #1340 .
- [x] Build and use on Windows (through a conda environment only, with ViSP installed in the conda environment) 
- [x] Dictate a clear way to install pybind or add it to third parties (maybe as a git module) ? See [the pybind doc](https://pybind11.readthedocs.io/en/stable/installing.html) => See the pr #1312 : In Virtualenv, specifying pybind is problematic, but pybind installation with conda is good.
- [x] Improve reporting in the configuration log for the reason why the Python bindings are not used. 
- [x] Create a single cmake target to run tests
- [x] Support for n-ary operator()
- [x] Increased the number of generated methods in the core module, clean up function parameters where required.


rolalaro:
You did a really great job, thanks a lot !!! :-)

PR #1359, opened 2024-03-27
[WIP] New photometric VS features + improvements to vpFeatureLuminance

SamFlt:
This PR introduces new visual servoing features, taken from the papers of Eric Marchand and readapted from his code.

These features are PCA and DCT compressed representations of the luminance information.

In terms of implementation, new classes are introduced:

- `vpFeatureLuminanceMapping` a visual servoing feature, which combines a `vpFeatureLuminance` object with a `vpLuminanceMapping` subobject. This is the feature that is given to `vpServo`
- `vpLuminanceMapping`, a common interface for mappings that operate on luminance information. a `map` method computes a vector representation from the image.  The `inverse` method, converts this vector back to an image, with some information loss. Finally, the `interaction` method computes the Jacobian, given the image, the compressed representation and the photometric Jacobian (computed beforehand by `vpFeatureLuminance` when used with the `vpFeatureLuminanceMapping`)
- The `vpLuminanceMapping` interface has two implementations: `vpLuminancePCA` and `vpLuminanceDCT`. they can be used interchangeably in `vpFeatureLuminanceMapping`

Before this PR is complete, several things are lacking:

- [ ] Encapsulate a Levenberg-Marquardt control law in a custom class.
- [ ] Add weighting mask when computing the error in `vpFeatureLuminance`
- [ ] Add an option to have a depth per pixel instead of considering that all 3D points are at an equal distance from the camera.


SamFlt:
@fspindle,

If the CI is ok, this PR can be merged. I will write a followup PR to implement the remaining features when I have some time :)

PR #1371, opened 2024-04-14
Fix vpRealSense2::getProductLine()

s-trinh:
Following error occurs when running example/device/framegrabber/grabRealSense2.cpp code with the D455 sensor:
- `Frame didn't arrive within 15000`

Tested with:
- `./grabRealSense2`
- `./testRealSense2_D435`
- `./testRealSense2_D435_opencv`

More info:
- Intel RealSense D455
- Firmware Version: 5.12.7.100
- Use Realsense2: yes (ver 2.54.2)

The error `Frame didn't arrive within 15000` still happens sometimes and relaunching the executable or unplug/plug the device seems to "solve" the issue.

### DIFF
--- a/modules/sensor/include/visp3/sensor/vpRealSense2.h
+++ b/modules/sensor/include/visp3/sensor/vpRealSense2.h
@@ -363,6 +363,8 @@ public:
 
   std::string getProductLine();
 
+  std::string getSensorInfo();
+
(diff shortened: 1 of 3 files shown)


fspindle:
@s-trinh Thanks again ;-)

PR #1396, opened 2024-05-06
Add basic I/O tests for visp::cnpy::npz_load(), visp::cnpy::npz_save() functions

s-trinh:
- [X] add `modules/core/test/tools/io/testNPZ.cpp` --> CI should fail
- [x] fix `parse_npy_header: failed to find header keyword: 'fortran_order'` error


fspindle:
@s-trinh Thanks. I find that not all the features you have introduced in the cnpy namespace are visible. What do you think about introducing a dedicated tutorial on https://visp-doc.inria.fr/doxygen/visp-daily/tutorial_python.html? Could you initiate something?

s-trinh:
@fspindle 
Good idea. I will see what I could do.

PR #1405, opened 2024-05-17
Fix compilation error related to GUI

rolalaro:
- Added a vpDisplayFactory namespace that permits to either get a newly allocated vpDisplay* if a GUI library is available or a nullptr otherwise. It permits to check only VISP_HAVE_DISPLAY instead of several macro in a software to get a new display.
- Fix compilation errors when no GUI library is available

### DIFF
--- a/modules/gui/include/visp3/gui/vpDisplayFactory.h
+++ b/modules/gui/include/visp3/gui/vpDisplayFactory.h
@@ -30,8 +30,10 @@
  *
  * Description:
  * Display Factory
- *
-*****************************************************************************/
+ */
+
+#ifndef vpDisplayFactory_h
+#define vpDisplayFactory_h
 
 #include <visp3/core/vpConfig.h>
 #include <visp3/gui/vpDisplayD3D.h>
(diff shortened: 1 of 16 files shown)


modules/gui/include/visp3/gui/vpDisplayFactory.h
```diff
@@ -0,0 +1,100 @@
...
+  return new vpDisplayD3D();
+#elif defined(VISP_HAVE_GDI)
+  return new vpDisplayGDI();
+#elif defined(VISP_HAVE_GTK)
+  return new vpDisplayGTK();
+#elif defined(HAVE_OPENCV_HIGHGUI)
+  return new vpDisplayOpenCV();
+#endif
+#else
+  return nullptr;
```
s-trinh, line 67:
```suggestion
  return NULL;
```

If we need compatibility with < C++11, otherwise maybe using things like `unique_ptr` to not have to deal with `delete`.

fspindle: @s-trinh To keep compat with c++9!, in `vpConfig.h` we added an emulation of `nullptr`. `NULL` has no more to be used

PR #1416, opened 2024-05-28
Nix packaging

nim65s:
Hello !

This PR includes:

1. a CMake fix for use cases where users (or package managers) set eg. `-DCMAKE_INSTALL_LIBDIR=/an/absolute/path`
2. a default.nix file which create a `visp` package in nix, and a flake.nix file to easily build and use this package. At some point, we will want to move this default.nix to https://github.com/NixOS/nixpkgs/, and just override the `src` attribute to use the local version instead.
3. a github action to test this nix setup.

I have a few questions:
- did I miss important dependencies / options ?
- visp do not compile correctly with ogre v14.2.5. Is this known ? What should we do about it ?
- contributing guidelines provide a "Copyright Assignment Agreement for an Individual Contributor". I am not in an individual position, but with CNRS. What should we do about that ?

Also, nix macos packaging is not done yet. I'll probably be able to find some time to deal with this, but not today, and maybe not this week.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,17 @@
+name: "Nix CI"
+
+on: [pull_request, push]
+
+jobs:
+  tests:
+    name: "Nix build on ${{ matrix.os }}"
+    runs-on: "${{ matrix.os }}-latest"
+    strategy:
+      matrix:
+        os: [ubuntu] # TODO: macos
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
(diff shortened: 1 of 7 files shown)


fspindle:
@nim65s Thanks for this nice contribution.

> did I miss important dependencies / options ?

I added other deps in the code review

> visp do not compile correctly with ogre v14.2.5. Is this known ? What should we do about it ?

No. I need to check.

> contributing guidelines provide a "Copyright Assignment Agreement for an Individual Contributor". I am not in an individual position, but with CNRS. What should we do about that ?

I don't know. I will investigate and let you know...

fspindle:
@nim65s Thanks Guilhem.

PR #1419, opened 2024-06-19
Particle Filter

rolalaro:
- [X] Mono-threaded version
- [X] Multi-threaded version
- [X] Example
- [X] Unit tests
- [X] Make the class template to use other classes than vpColVector as measurements
- [ ] Tutorial: code
- [ ] Tutorial: documentation


rolalaro:
@fspindle Should I make the vpParticleFilter class a template class to allow
having other types than vpColVector for the measurements ?
It would affect the signature of the vpLikelihoodFunction but would have no other
effects on the class.
It would permit to use e.g. a pcl::PointCloud for the measurement instead of having to convert it
into a vpColVector. However, as far as I know, template classes are not allowed in Python, so the 
Python wrapping would be limited to vpColVector.

PR #1434, opened 2024-07-03
Fixing ubuntu_dep_src CI

rolalaro:
Use a function to get the most recent tag of the 3rdparty git repositories to use them instead of dev branches


fspindle:
@rolalaro Nice changes to retrieve the latest version

PR #1437, opened 2024-07-04
Fixing CI ubuntu-dep-src

rolalaro:
Fix attempt on ci ubuntu dep src


fspindle:
@rolalaro For information as given [here](https://vsupalov.com/docker-arg-env-variable-guide/#arg-and-env) both syntaxes to set `ENV ` variable are valid:
```
ENV foo /bar
# you can also write
ENV foo=/bar
```

PR #1445, opened 2024-07-22
Upgrade to apriltag 3.4.2

fspindle:


fspindle:
I will close this PR.
There is a better onr in PR #1448 that brings less changes.

PR #1456, opened 2024-08-28
Modifications for the upcoming Render-Based Tracking submodule

SamFlt:
This PR introduces small modifications and improvements that were required for the Render-Based Tracking submodule that is coming soon (hopefully :-)). It contains:

- Changes to `vpMeSite`:
   -  Introduction of a new method, `trackMultipleHypotheses`, that retains the K best candidates instead of only one.
   -  Small modifications to the `track` code to cleanup code redundancy, removed the use of an array that was not needed.
- Some improvements to the rendering pipeline
  - Some optimizations in the normal pipeline (although pretty insignificant). 
  - Cleaned some code and refactored some variables
  - Added the auxiliary methods to render, that are called before and after rendering. This allows for inheriting classes to further process data before rendering.
  - Improvements to the tutorial code and documentation.
  - Cleared transform on loaded models: this led to inconsistencies, as BAM files saved the Object-to-world transform.
  - Better clipping value computation.


modules/gui/include/visp3/gui/vpDisplayFactory.h
```diff
@@ -199,8 +200,100 @@ std::shared_ptr<vpDisplay> createDisplay(vpImage<T> &I, const int winx = -1, con
...
+    throw vpException(vpException::dimensionError, "Too few images for the grid size");
+  }
+  (void)res;
+  (void)settings;
+  (void)currRow;
+  (void)currCol;
+  (void)currentPixelX;
+  (void)currentPixelY;
+  (void)maxRowHeightPixel;
+}
```
fspindle, line 231: This function seams useless or not fully implemented

modules/ar/include/visp3/ar/vpPanda3DPostProcessFilter.h
```diff
@@ -62,13 +62,27 @@ class VISP_EXPORT vpPanda3DPostProcessFilter : public vpPanda3DBaseRenderer
     m_renderOrder = m_inputRenderer->getRenderOrder() + 1;
   }
 
+  ~vpPanda3DPostProcessFilter()
+  {
+    // delete m_texture;
+    // delete m_buffer;
+  }
```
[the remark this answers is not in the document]
SamFlt: I will remove them.
I was under the impression that Panda3D managed the texture and graphics buffer (through the WindowFramework). after having another look at the examples, I'm not so sure. I will try to replace the raw pointers with Panda's managed memory system (using `PointerTo`), so that there are no clashes with panda potentially releasing the memory by itself.

modules/gui/include/visp3/gui/vpDisplayFactory.h
SamFlt: This function is the end condition of the recursion of the function. Since there are no vpImage passed as an argument, we consider that we have iterated over all the displays to create. If the dimensions are incorrect (currCol vs cols and currRow vs rows) then we raise an exception. Otherwise, res is filled with all the created displays, and `makeDisplayGrid` can return the result.

fspindle:
@SamFlt Thanks

PR #1463, opened 2024-09-13
Add mid-optimization normal filtering to MBT

SamFlt:
This PR adds face rejection mid optimization for the MBT for the normal tracker. This is necessary, as filtering faces before VVS is not stringent enough: some faces that are close to being rejected before optimization have a strong adverse effect on VVS, leading to divergence. This is due to the fact that as the plane gets closer to being parallel to the camera axis and D gets to 0, the interaction matrix blows up.

This can observed in the Blender MBT tutorial, where turning on normals led to divergence before the fix.


fspindle:
@SamFlt thanks

PR #1465, opened 2024-09-15
Add RISC-V architecture in GitHub CI

s-trinh:
Add RISC-V architecture. There should be more development in the future with this architecture, especially from China due to potentially licensing conflict with Arm, e.g.
- https://github.com/opencv/opencv/wiki/OpenCV-RISC-V

---

Suggestion to use `ubuntu-latest`.

Currently `ubuntu-latest` == `ubuntu-22.04`:
- [Available images](https://github.com/actions/runner-images)

image

- [Standard GitHub-hosted runners for public repositories](https://docs.github.com/fr/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

image

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  build-coverage:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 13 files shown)


fspindle:
@s-trinh Thanks

PR #1491, opened 2024-10-31
Fix broken conda linux CI

olivier-roussel:
It seems that last libxml2 conda package release used is a variant without `icu` library, whereas it actually needs `icu`.


fspindle:
@olivier-roussel Thanks

PR #1501, opened 2024-11-06
Fix warnings and issue in tutorial/imgproc/tutorial-circle-hough.cpp

fspindle:


fspindle:
@rolalaro For information

PR #1511, opened 2024-11-28
Canny stack

rolalaro:
- Deactivated the stack check and potential change by default, it is now an option
- Corrected a compilation warning intutorial-canny


fspindle:
@rolalaro Perfect. Thanks

PR #1516, opened 2024-12-03
OpenCV display refacto

rolalaro:
[CORE] Created an implementation class for the OpenCV-based display in order to remove the inclusion of opencv in the header


fspindle:
@rolalaro Thanks. Perfect

PR #1520, opened 2024-12-10
Add xorg-xorgproto dep on linux

olivier-roussel:
Fixes broken Conda-based CI on linux, most probably due to an update of a conda package.


fspindle:
@olivier-roussel Thanks for the fix

PR #1526, opened 2024-12-19
Compatibility with Ogre 1.12

rolalaro:
Updated vpAROgre and visp_ar CMakeLists to make ViSP compatible with Ogre 1.12


fspindle:
@rolalaro When I run the example on my Ubuntu 24.04 with Ogre 1.12 installed from existing package, I have an exception 

```
$ ./AROgre -c
RenderSystem Name: OpenGL Rendering Subsystem

Catch an Ogre exception: Unable to open zip file '/usr/share/OGRE/Media/packs/oa_rpg3dm2.pk3'
```

```
$ ls /usr/share/OGRE/Media/packs/
profiler.zip  SdkTrays.zip
```

PR #1533, opened 2025-01-03
Update OpenBLAS include dir seach paths for static library

s-trinh:
When using `make -j$(nproc) install PREFIX=$(pwd)/install`:
- static and shared libraries are created: `libopenblas.a` and `libopenblas.so`
- `install/include` contains
```bash
cblas.h  f77blas.h  lapacke_config.h  lapacke.h  lapacke_mangling.h  lapacke_utils.h  lapack.h  openblas_config.h
```

When using `ccmake` instead, the `openblas` folder is created and only in static build `libopenblas.a`:
- `install/include/openblas`
```bash
cblas.h  lapacke_config.h  lapacke_example_aux.h  lapacke.h  lapacke_mangling.h  lapacke_utils.h  lapack.h  openblas_config.h
```


fspindle:
@s-trinh Thanks

Issue #1519, opened 2024-12-04
Dataset not detected when tests are disabled


rolalaro:
Hello,

I noticed something strange: if BUILD_TESTS is set to OFF, the visp-images dataset is
not looked for. It affects some of the examples (at least example/ogre-simulator/AROgre.cpp)
which cannot load correctly the mire-2 dataset.

Is it a desired feature ?

Best regards

fspindle:
No this is a bug

fspindle:
Fixed in PR #1538

PR #1538, opened 2025-01-08
Compatibility with more recent versions of Ogre

rolalaro:
- Fix compatibilities issues with Ogre 1.12.10, 13.1.0 and 14.0.0 
- Created 2 CIs to check the compilation of ViSP when Ogre is used as 3rd party. Unit tests for Ogre cannot be run because Ogre requires a valid X11 context


rolalaro:
@fspindle the bug should have been fixed in the latest commit, could you please pull and try again ?

fspindle:
Closes #1547 #1519 #1546

PR #1556, opened 2025-01-22
Add visp_vision module missing dependency to tutorials

fspindle:


fspindle:
Build issue occurs when `vision` module is disabled from build

PR #1560, opened 2025-01-24
Aarch64 CI compilation segfault fix

rolalaro:
[CI] Trying to fix segfault during compilation of the aarch64 CI


fspindle:
@rolalaro Can you echo the content of the `NB_PROC_TO_USE` var to understand what's wrong. Maybe, the solution would be to use `-j1`

PR #1627, opened 2025-03-18
Rough support of OpenCV ArUco fiducial markers

s-trinh:
It is possible to detect and decode OpenCV ArUco tags using the AprilTag 3 code.
Since ArUco markers are widely used due to OpenCV, it could be a good idea to support also the decoding of ArUco tags.

Pros:
- precision in the extraction of the corners location using the AprilTag pipeline

Cons:
- need a white border around the tag, contrary to using the OpenCV library
- no marker occlusion
- detection range? (I think AprilTag far detection range is lower than ArUco)

Marker families are similar to OpenCV: https://docs.opencv.org/4.10.0/de/d67/group__objdetect__aruco.html#gga4e13135a118f497c6172311d601ce00da395a8309569c1977538fd11b57752087 but without the `7x7` family.

Warning:
- there is OpenCV ArUco and the [official one](https://sourceforge.net/projects/aruco/), not sure how they perform between the two

---

TODO:
- [x] add tests for accuracy and regression about all the families
- [x] which license to use? --> I will add the AprilTag and the OpenCV licenses
- [x] understand why there is no detection with for instance `Aruco_6x6_1000-27.png` image
- [x] update the doc
- [x] scale the rectangle to indicate the first corner with the tag size
- [x] update MSVC warnings for the new files? see `3rdparty/apriltag/CMakeLists.txt`
- [x] understand why there are lots of false detections
- [x] check / recheck that AprilTag ArUco and OpenCV ArUco return the same convention for the corners location and the pose frame &#8594; pose frame is identical ([run this](https://github.com/opencv/opencv/blob/01ef38dcad65a119204e6d70f553767a7d2ab614/samples/cpp/tutorial_code/objectDetection/detect_markers.cpp)) but the order of the corners is not the same (bottom left for AprilTag and top left for OpenCV). Should not be an issue since most of the use cases are pose computation or tag id decoding.
- [x] propagate the `--decision-margin` option in the different tutorial code
- [x] return the decision margin for each detection with `getTagsDecisionMargin()`
- [x] add decision margin threshold setter `setAprilTagDecisionMarginThreshold()` to filter out false positive detection occuring mainly with AprilTag 16h5 and ArUco 4x4, 5x5 and 6x6 families. By default decision margin threshold is not used (current value is set to -1). When set to 50, there are less false detections, but some remain. Can be raised to 100 to be more reliable. When values are greater than 0, this threshold is applied to all families.
- [x] return the hamming distance for each detection with `getTagsHammingDistance()`. Values are in range 0-255  and corresponds to the number of corrected bits
- [x] add hamming distance threshold setter `getAprilTagHammingDistanceThreshold()`. By default the threshold is set to 0, meaning that only tags detected without bits correction are retained. It allows to filter out almost all the false positive detections. The decision margin threshold can than be used to filter out the few false positive detections that may remain for example with ArUco 4x4 family. This threshold is applied to all families.
- [x] current code in `tutorial/detection/tag/tutorial-apriltag-detector.cpp` parse the message to retrieve the tag_id, add a comment to tell that there is `detector.getTagsId()` doing it (or update the code?, unless we want to keep the parsing code as an example)
- [x] Update doxygen tutorial accordingly 

To generate quickly an ArUco:
- https://chev.me/arucogen/


fspindle:
I did some tests with tutorial-apriltag-detector-live.cpp
- ArUco 4x4 doesn't work well compared to OpenCV version. There are a lot of false positives
Screenshot from 2025-03-21 16-34-37
Screenshot from 2025-03-21 16-37-32
- ArUco 5x5 works better, but there remain some false detections, like in the next image where 2 tags are detected
Screenshot from 2025-03-21 16-39-49
- ArUco 6x6: there are fewer false detections than for ArUco 5x5, but some still remain

s-trinh:
Thanks for the tests.

---

There is a `decision_margin` in AprilTag code.
Previous behavior was to not filter the detections.
This can be used to filter the false detections with ArUco 4x4, 5x5 and 6x6. Also, the preferred ArUco family is `MIP_36h12`, [cf](https://github.com/clearpathrobotics/aruco):

<blockquote>
The main difference with previous library is that you can select amongst a set of Dictionaries:   ARUCO_MIP_25h7,ARUCO_MIP_16h3, and ARUCO_MIP_36h12 which are hardcoded. Thus, you do not need external files describing the ditionary.

We recommend using ARUCO_MIP_36h12, however, for compatibilty with old users, the default dictionary of the library is ARUCO, which is the original dictionary in previuos versions.  However, we recommend using ARUCO_MIP_36h12 because it is more robust to errors.
</blockquote>

Currently, this `decision_margin` is arbitrary set to 100. This could degrade performance (range detection) with AprilTag.
Since the recommended ArUco family is `MIP_36h12` which does not suffer from these lots of false positives, another option is to set it to a way lower value (e.g. 20 or something).
On another side, I don't know which families are usually used with OpenCV ArUco, with regular users. And I don't know the correspondences between OpenCV 4x4, 5x5, 6x6 with the official ArUco, if there are completely outdated for instance.

In any case, there is a method to change this value.
OpenCV `example_tutorial_detect_markers` sample to test the ArUco detection does not suffer from these high false detections. Could be inherent to how AprilTag and ArUco work or could be that OpenCV use a threshold to filter the detections.

fspindle:
@s-trinh If ci doesn't point any issue, on my side I'm ready to merge

s-trinh:
As a first ArUco support, I would be more conservative and try to not potentially touch/degrade the AprilTag performance with default values.

From what I have understood:
- Hamming distance is for bit-error correction, so a default value of 0 looks too conservative
- the decision margin should be the parameter to be adjusted, for instance:
  - https://github.com/AprilRobotics/apriltag/issues/361
  - https://github.com/AprilRobotics/apriltag/issues/375

I don't know why with the ArUco pattern there are so many false detections.

If we do not want to have the decision margin being applied only for the 4x4, 5x5 and 6x6 ArUco patterns, one possible solution would be to just use a margin of 0 (no filtering) and warn the user in the doc. Since this parameter can be easily modified using the tutorial code, it should be fine.

Also, the recommended ArUco family in the OpenCV doc is MIP_36h12 which does not have all these false detections.

fspindle:
@s-trinh You are right, we should be conservative to keep the same behavior as in the previous releases.
- Default hamming distance threshold is set by default to 2 in `vpDetectorAprilTag` class
- Default decision margin threshold is set by default to -1 (meaning not used)  in `vpDetectorAprilTag` class

s-trinh:
I have updated the doc to mention that `setAprilTagDecisionMarginThreshold()` or `getTagsDecisionMargin()` can be used to filter the detections.

See https://stackoverflow.com/a/51511558 for a reference about `ARUCO_MIP_36h12` being the recommended ArUco family.

fspindle:
@s-trinh Thanks for this nice contrib

PR #1640, opened 2025-04-01
Fix catchRBT segfault

SamFlt:


modules/tracker/rbt/test/catchRBT.cpp
```diff
@@ -597,7 +597,7 @@ SCENARIO("Instantiating a render-based tracker", "[rbt]")
       REQUIRE((silset.preferPreviousPoints()));
 
       REQUIRE((tracker.getOptimizationGain() == 1.0 && tracker.getMaxOptimizationIters() == 10));
-      REQUIRE((tracker.getOptimizationInitialMu() == 0.5 && tracker.getOptimizationMuIterFactor() == 0.1));
+      REQUIRE((tracker.getOptimizationInitialMu() == 0.5 && tracker.getOptimizationMuIterFactor() == 0));
```
SamFlt, line 600: That's weird: on my machine, the test passes with the value 0.1 but not 0. This is the expected behaviour, as "muIterFactor is set to 0.1 in the json settings (at line 536 of catchRBT.cpp)
There was an issue on master that led to this result, but it was fixed in [9ffe617](https://github.com/lagadic/visp/pull/1640/commits/9ffe6170b55860052cc16de67fb3a480b1c0e0f6).

fspindle: It was due to my `LD_LIBRARY_PATH` pointing to an other version of visp libraries.
I'm now able to get `tracker.getOptimizationMuIterFactor() == 0.1`

PR #1651, opened 2025-04-09
Fix rbt build in conda env

fspindle:
Following changes allow to build visp python bindings using a conda environment on macos.


fspindle:
@SamFlt For information

PR #1658, opened 2025-05-04
tests segfaults on Mac intel

nim65s:
Hello,

As found in https://github.com/NixOS/nixpkgs/pull/403603#issuecomment-2849001986, 57 tests out of 120 segfault on macos intel with nix.

This PR update the nix CI to test arm+intel / mac+ubuntu, to show it is still the case on master.
It probably should no be merged before those segfaults are sorted out :)

Or maybe this is already known, and mac intel platforms are considered out of scope / obsolete ?

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -11,10 +11,10 @@ concurrency:
 jobs:
   tests:
     name: "Nix build on ${{ matrix.os }}"
-    runs-on: "${{ matrix.os }}-latest"
+    runs-on: "${{ matrix.os }}"
     strategy:
       matrix:
-        os: [ubuntu, macos]
+        os: ["ubuntu-24.04", "ubuntu-24.04-arm", "macos-13", "macos-14"]
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


fspindle:
@nim65s 

> Or maybe this is already known, and mac intel platforms are considered out of scope / obsolete ?

I was not aware of these issues, and mac intel platforms remain supported.

Could you tell me how to reproduce this error on a local intel mac?

nim65s:
Ok, thanks for the test. 
Let's try nix with clang 15 :)

nim65s:
NB: visp is build with clang 15 here, but not its dependencies

nim65s:
The CI show the same issue with
```
visp> --   C/C++:
visp> --     Built as dynamic libs?:      yes
visp> --     C++ Compiler:                /nix/store/q7yahgxdah0aawgy6ydvpsaqmxykminy-clang-wrapper-15.0.7/bin/clang++  (ver 15.0.7)
```

Therefore, I guess this is an issue in a dependency as packaged by nix, and not in visp, so this PR can probably be closed.

Thanks for the feedback !

PR #1659, opened 2025-05-04
nix: enable ogre

nim65s:
Hi again,

It seems that ogre is now working with nix, so we can enable it :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -420,23 +420,6 @@ if(OGRE_DIR)
   STRING(REGEX REPLACE "\\\\" "/" OGRE_DIR ${OGRE_DIR})
 endif()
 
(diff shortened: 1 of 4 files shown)


fspindle:
@nim65s Thanks for this PR.  

Concerning the error on Darwin, it seems that `FindPkgMacros.cmake` CMake file that is included in `OgreTools.cmake` is not found on Darwin. This file defines the missing macro `get_debug_names()`.

How can I enable Darwin to test and fix this issue ?

fspindle:
@nim65s Issue you reported should be fixed now

nim65s:
Thanks, and sorry for not taking the time to answer earlier :)
To enable ogre on darwin, `ogre` needs to be in `buildInputs`, and no only when `stdenv.isLinux`

PR #1684, opened 2025-05-20
Fix pcl potential runtime error

fspindle:


s-trinh:
FYI, some info I have asked in the PCL repo:
https://github.com/PointCloudLibrary/pcl/issues/5806#issuecomment-2900397246

fspindle:
@s-trinh Thanks, I didn't see that before. I'm now using `INTERFACE_COMPILE_OPTIONS` property that seems better than using `PCL_COMPILE_OPTIONS`

PR #1736, opened 2025-07-19
Add a quick tutorial about automatic gamma correction

s-trinh:
The tutorial shows the code to use to compare the results between the different auto gamma correction methods.


fspindle:
@s-trinh Thanks

PR #1744, opened 2025-07-31
Handling vpImage<float> in vpColorDepthConversion

rolalaro:
[CORE] vpColorDepthConversion now handles vpImage<flaot>

### DIFF
--- a/modules/core/src/camera/vpColorDepthConversion.cpp
+++ b/modules/core/src/camera/vpColorDepthConversion.cpp
@@ -40,7 +40,7 @@
 
 // System
 #include <algorithm>
-#if (VISP_CXX_STANDARD > VISP_CXX_STANDARD_98)
+#if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_17)
 #include <optional>
 #endif
 


modules/core/src/camera/vpColorDepthConversion.cpp
```diff
@@ -40,6 +40,9 @@
 
 // System
 #include <algorithm>
+#if (VISP_CXX_STANDARD > VISP_CXX_STANDARD_98)
```
fspindle, line 43: `std::optional` is only available since c++17.
Since here, we are in a `.cpp`, the test should be `#if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_17)`

modules/core/src/camera/vpColorDepthConversion.cpp
```diff
@@ -134,32 +137,16 @@ vpColVector deproject(const vpCameraParameters &intrinsic_cam_params, const vpIm
 #endif
 }
 
-} // namespace
-
-#endif // DOXYGEN_SHOULD_SKIP_THIS
+#if (VISP_CXX_STANDARD >= VISP_CXX_STANDARD_17)
```
fspindle, line 140: ok

modules/core/src/camera/vpColorDepthConversion.cpp
rolalaro: My bad, it was a typo. FIxed in [f3f3507](https://github.com/lagadic/visp/pull/1744/commits/f3f35070930b5333186895e3b7339c9d87928651)

PR #1747, opened 2025-08-11
Fixing codecov CI

rolalaro:


rolalaro:
[Link related to the fix of the testDisplayRoi)](https://stackoverflow.com/a/10987416 ) : they talk about a memory leak when the option `-noreset`  is not used, so I tried to add it to see if it fixes the test

fspindle:
@rolalaro Thanks

PR #1772, opened 2025-09-20
Use memcpy in vpImageConvert::YUV420ToGrey()

s-trinh:
- call to `setAprilTagHammingDistanceThreshold()` seems to be missing in some files
- when looking at the `vpImageConvert::YUV420ToGrey()` source code, the for loop can be replace by a `memcpy`


fspindle:
@SamFlt Well spotted, thanks

Issue #1773, opened 2025-09-21
Java binding : VpDetectorAprilTag::getTagsCorners() does not work with Android


s-trinh:
Calling `VpDetectorAprilTag::getTagsCorners()` from Java and Android project does not seem to work with the following:

https://github.com/lagadic/visp/blob/af3e8a3918dc60e932053c6b926c3ee0a77a31fe/modules/java/misc/detection/gen_dict.json#L198-L214

It works with:

https://github.com/lagadic/visp/blob/e3afa2a50200f6aecd7395859f97a0660e2d435d/modules/java/misc/detection/gen_dict.json#L198-L214

---

It works correctly with pure Java code, see https://github.com/lagadic/visp/blob/af3e8a3918dc60e932053c6b926c3ee0a77a31fe/tutorial/java/tag-detection/ApriltagDetection.java

Tested with:

```bash
openjdk 21.0.8 2025-07-15
OpenJDK Runtime Environment (build 21.0.8+9-Ubuntu-0ubuntu124.04.1)
OpenJDK 64-Bit Server VM (build 21.0.8+9-Ubuntu-0ubuntu124.04.1, mixed mode, sharing)
```

PR #1774, opened 2025-09-21
Update the Android AprilTag app

s-trinh:
I had to update the `gen_dict.json` file to be able to use `detector.getTagsCorners();`, need to double check why it was needed:
- see the commit https://github.com/lagadic/visp/commit/e3afa2a50200f6aecd7395859f97a0660e2d435d
- corresponding issue: https://github.com/lagadic/visp/issues/1773
- why the fix is needed, the Java code seems to work before: https://github.com/lagadic/visp/blob/af3e8a3918dc60e932053c6b926c3ee0a77a31fe/tutorial/java/tag-detection/ApriltagDetection.java#L150-L160

---

Video captures:

- should be quad_decimate=2, nthreads=1 

https://github.com/user-attachments/assets/1380a254-abaf-4ed2-ac03-5fda6a4223e5

- should be quad_decimate=4, nthreads=4

https://github.com/user-attachments/assets/2bf01bd5-6cf5-4be0-a591-a8edb04adfff 

---

The settings panel:

- the camera intrinsic parameters are automatically computed from the hardware specs
- don't know what happens if these info are not available, if the app crashes
- with autofocus, the computed focal length can be slightly off

---

The issues are:
- flickering display / drawing of the primitives, most likely a separated thread for drawing is needed
- the app will crash after a certain amount of time
- some possible memory leak
- some code come from ChatGPT, so not sure if it is the correct approach wrt. the Android environment
- the state machines process behind is not totally handled correctly for the camera:
  - corresponding error: "java.lang.RuntimeException: Camera is being used after Camera.release() was called"
  - same when resuming the app, etc.
- handling of the landscape mode: the image is distorted
- the current code uses the obsolete Camera API (instead of Camera2)
- ...

---

App logo after tweaking the results from ChatGPT:

|   |  |
| -------- | ------- |
|  |   |

### DIFF
--- a/doc/tutorial/android/tutorial-android-getting-started.dox
+++ b/doc/tutorial/android/tutorial-android-getting-started.dox
@@ -69,7 +69,7 @@ Open Android Studio.
   sources are located in the `sdk` folder. As shown in the next image, you should add the following lines:
 \code
 include ':visp'
-project(':visp').projectDir = new File('../sdk')
+project(':visp').projectDir = new File('sdk')
 \endcode
 \image html 05-editing-global.jpg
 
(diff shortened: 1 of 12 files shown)


fspindle:
@s-trinh Thanks

PR #1775, opened 2025-09-23
Add workflow for staled issues

olivier-roussel:
As discussed in our last dev meeting, this workflow will stale issues according to the following policy:
- Mark as "staled" and comment any issue that had no activity for 30 days
- Comment and close any issue that had no activity for 30 days since it has been marked as "staled"

This GH action has also features for PRs, but they are not enabled yet.

### DIFF
--- a/.github/workflows/stale-github-issues.yml
+++ b/.github/workflows/stale-github-issues.yml
@@ -1,0 +1,23 @@
+name: Stale inactive issues
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


olivier-roussel:
Tested on private repository here: https://github.com/olivier-roussel/test-stale-issues .
See issue status evolution example here (delays have been reduced to 0 for testing): 
https://github.com/olivier-roussel/test-stale-issues/issues/1

PR #1778, opened 2025-09-24
RBT Various additions and improvements

SamFlt:
This PR introduces new functionalities to the Render-Based Tracker:

General:
- [x] The render data now contains more information, such as the object's diameter (bounding box extent) and 3D corners in object space.
- [x] The tracker can now be reset and the model reloaded more easily, without bugs related to Panda3D appearing.  The reset function should be called when changing object or reinitalizing tracking. Some of the tracker's parts are stateful so this is important.
- [x] Rendering is now done at a lower precision. High FP32 precision is not really needed, and GPU to CPU transfer is extremely slow. For the Geometry renderer, The amount of transmitted bytes has been divided by 4. These changes had no significant impact when evaluating on several benchmarking datasets. In my tests, they improved render speed by around ~30%. For instance, a render that took 7ms now takes 5, which is not negligible when considering that rendering can happen twice per frame (when odometry is activated).
- [x] Reworked the convergence metric (early stopping for the optimizer and rerendering criterion). Added the reprojection error as a criterion, which is easier to reason with.
- [x] Fixed a bug where the object contours that were too close to the computed object's 2D bounding box were not detected.
- [x] Update documentation with new features and odometry. 

Masking:
- [x] a Rework of the color masking process to use logistic regression instead of the previous solution that was based on the object/background probability ratio along with an hyperparameter threshold to be tweaked by the user. The masks are now generally far less noisy. An another solution instead of logistic regression would be to use the Bayes'rule, but this solution fails when we consider a small object that has a color similar to that of the background.
- [x] Some computational speedup in color masking.  
- [x] Introduced a new Depth mask strategy
- [x] Introduced a Combined color/depth masking strategy. The strategy is to take the minimum of the probabilities estimated by the color or depth mask.

Features:
- [x] Introduced an experimental  trackable feature, based on photometric Visual servoing.
- [x] In Python, added XFeat keypoints as trackable features.
- [x] Some of the M-estimators now use adaptive values that depend on the object's size in the image or in 3D space
- [x] Improved speed for the depth based tracker

Odometry:
- [x] Integrated the XFeat Visual Odometry strategy.

Pose estimation:
- [x] Developed two pose estimation approaches based on leveraging XFeat to solve the PnP problem.  
- [x] Usage tutorial and script to use those methods

Others:
- [x] Integrate Python JSON parsing for XFeat tracker and odometry

Known issues:
- As there is now a more significant part of the code that is pure Python, it has become apparent that there is an issue with the stubs and documentation generation: The Python part is not parsed and Sphinx (used for the doc) does not correctly generate the pages for each class. This will be the subject of a later PR. EDIT: this is partially fixed.

### DIFF
--- a/modules/tracker/rbt/src/mask/vpColorHistogramMask.cpp
+++ b/modules/tracker/rbt/src/mask/vpColorHistogramMask.cpp
@@ -143,7 +143,9 @@ void vpColorHistogramMask::updateMask(const vpRBFeatureTrackerInput &frame,
 
   if (m_computeOnBBOnly) {
     mask.resize(height, width, 0.f);
+#ifdef VISP_HAVE_OPENMP
 #pragma omp parallel for
+#endif
     for (int i = top; i <= bottom; ++i) {
       for (int j = left; j <= right; ++j) {
         unsigned int index = m_histObject.colorToIndex(frame.IRGB[i][j]);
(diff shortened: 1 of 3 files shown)


SamFlt:
@fspindle I think this PR is ready to be merged.

PR #1779, opened 2025-09-25
Python improvements for apriltag class

SamFlt:
This PR improves the Bindings for the Apriltag detection class.

It also introduces a script that can be used to perform apriltag detection on a sequence (either a stored file sequence or through a realsense camera).  Optionally, It stores the detection results in a json file

By default, it connects to a Realsense camera (requires the pyrealsense2 package)

After installing the bindings, it can be run by simply typing:
```
$ visp-python-apriltag-detection --log-path ~/log.json
```


fspindle:
@SamFlt Thanks

PR #1783, opened 2025-10-03
Python stubs mypy

SamFlt:
This PR improves the stub generation process:

* [x] import visp.core is now correctly resolves by IDEs
* [x] Replaced pybind11_stubgen by mypy, which is a larger scale typing project. This allows to generate stubs for Python source files and should also be better supported.
* [ ] Implement build isolation, so that mypy is not installed in the user's environment.


SamFlt:
@fspindle if you're okay with it, this PR can be merged.

PR #1785, opened 2025-10-07
Add coverage task with gcovr, add coverage CI

SamFlt:
Experimenting with migrating coverage workflow to a pixi environment.


SamFlt:
@fspindle  Ready for merging.

PR #1810, opened 2025-10-30
Revert macOS intel runner from v15 to v13 for pixi CI

olivier-roussel:
All our pixi based CI on macOS for intel are failing at the subs generation phase (we do not know the reason yet).
This appears only on macOS intel (i.e. x64) runners since the bump to v15. The CI seems fine with macOS v13. The CI is also fine with macOS v15 arm64. 
The bump to macOS v15 intel was motivated by the close stop of support for macOS v13 runners from GH ( https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/ ), with interruptions of service starting on 4th november and shut down one month later.
We did not spot this on the PR that included the bump ( #1805 ) because there was another error in the python pinning version scheme (making use of *compatible versions* instead of exact MAJOR.MINOR versions), and for some reason, the current compilation error on macOS v15 intel disappear when some part of ViSP are compiled with Python 3.14.

I also add in this PR the switch of channel for the panda3d conda packages: we moved our forge from Anaconda servers to Prefix.dev servers as Anaconda servers lack of disk space for new rebuild.

This PR should be backported to all PRs opened after #1805 merge.


olivier-roussel:
@fspindle If you are OK with that, I suggest we merge this as soon the last CI is green, so we can backport it the other concerned PRs (even if some already have most of the changes) before the end of the week.

PR #1825, opened 2025-11-17
Fix mac-15-intel support

olivier-roussel:
CI was failing on MacOS 15 intel with Pixi due to a opencv package with broken symbols (https://github.com/conda-forge/opencv-feedstock/issues/493), due to some problems with macOS SDK v14 if I understood correctly. This has been fixed in recent concerned opencv packages (osx-64,  v4.12.0 with build number >= 607). However, this recent opencv packages are not compatible yet with ogre current package due to some common dependencies rebuild in progress that are stalling (see https://github.com/conda-forge/freeimage-feedstock/pull/54 and https://github.com/conda-forge/freeimage-feedstock/pull/55, at least).
Workaround for now is to disable ogre support on osx-64.


fspindle:
@olivier-roussel Seems ready to merge to me

olivier-roussel:
@fspindle Not yet, I have fixed the freeimage conda package (https://github.com/conda-forge/freeimage-feedstock/pull/56 and https://github.com/conda-forge/freeimage-feedstock/pull/57) that was preventing ogre to be compatible with last opencv package. But yesterday evening the package was still not available despite it was already uploaded to conda-forge channel (it requires some time to sync, maybe affected by general shutdown).
I'd like to try to re-enable ogre also on osx-64 so, since we've found the problem.

olivier-roussel:
Last ogre conda package rebuild is now available (https://github.com/conda-forge/ogre-feedstock/pull/144), which is compatible with fixed opencv package for osx-64.
I just re-enabled ogre on all platform so.
One consequence of all that is we need to bump our python minimal version supported for Pixi from 3.10 to 3.11 as last numpy 2 packages are not rebuild anymore with python 3.10.

fspindle:
@olivier-roussel and @SamFlt I see no reason not to upgrade to Python 3.11 as the minimum required version.

olivier-roussel:
@fspindle  In the meantime I tried to fix the linking error we got with Ogre and Visp on windows, when using recent Ogre versions (>v1.12, so starting from v13 as they changed their versionning scheme at this point). 
On MSVC, the linking error was quite weird and something like:
```
D:\a\visp\visp\.pixi\envs\full-ci-python-supported\Library\bin\OgreBites.dll : fatal error LNK1107: invalid or corrupt file: cannot read at 0x318
```
(see https://github.com/lagadic/visp/actions/runs/19507760178/job/55846719177#step:5:1303).
After working on a minimal reprocudible example, it appeared that [4095ce4](https://github.com/lagadic/visp/pull/1825/commits/4095ce46971808a8854c5aa1be176723e4217b4a) was causing the error. But this fix revealed a deeper bug on Ogre CMake custom macros. It appears that we don't need our custom CMake code that we do in `modules\ar\CMakeLists.txt` to create customized targets and filter others with modern (>= v13) Ogre version, we can just use their CMake vars as it and everything looks fine. So [297a1df](https://github.com/lagadic/visp/pull/1825/commits/297a1df7c3325031f9263af6fcb222673916ee98) bypass all this cmake code for >=v13 Ogre. 

So now Ogre should work again with Pixi on Windows, let's wait the CI...

fspindle:
@olivier-roussel Thanks for investigating that. I will do additional tests on my side

fspindle:
@olivier-roussel I will completely refactor Ogre usage in `module/ar`. I have something working on Ubuntu. I will also test it ASAP under windows before a commit

fspindle:
@olivier-roussel Seems ready to merge

olivier-roussel:
@fspindle Looks good to me. That's great, it's far more readable and clean like this.

PR #1835, opened 2025-11-24
Add support for ArUco 7x7 tag family

s-trinh:
ArUco `7x7` have bigger [Hamming distance](https://docs.opencv.org/4.12.0/de/d67/group__objdetect__aruco.html#gga4e13135a118f497c6172311d601ce00dada8e830ff0024e839e93c01f5fed0c55) between two codes and the id decoding should be more reliable.

Support for ArUco 4x4, 5x5 and 6x6 was much easier since we can use `tag16h5`, `tag25h7`, `tag36h11` as a starting point.

### "bit-pattern" 

First compute the correct "bit-pattern" for the considered family, see: [tf->bit_x / tf->bit_y](https://github.com/AprilRobotics/apriltag/blob/22b44c13c6f0b68d7eaa1504cec0b8a20850122e/tag36h11.c#L628-L631)

With fiducial marker detection, it goes roughly as "quad_rectangle extraction" --> "perspective correction using estimated homography" --> "pattern decoding".

In OpenCV, one tag id is stored using 4 values to cover the 4 orientations: https://github.com/opencv/opencv/blob/3c3596c6efcabd0e5ee021e2048a30ec38a8d4e1/modules/objdetect/src/aruco/predefined_dictionaries.hpp#L21-L25

With the AprilTag library, some kind of bitwise operations are used to cover the 4 tag orientations, saving some memory at the cost of extra cpu work.

image

For a `5x5` family, the previous image shows the `tf->bit_x / tf->bit_y` values to extract the tag pattern into a full codedata such that using bitwise operations it is possible to cover the 4 tag orientations. More [information here](https://github.com/AprilRobotics/apriltag/issues/374#issuecomment-2727489300).

### Codedata

Once the `tf->bit_x / tf->bit_y` values are computed for a `7x7` family, corresponding [`codedata`](https://github.com/AprilRobotics/apriltag/blob/22b44c13c6f0b68d7eaa1504cec0b8a20850122e/tag36h11.c#L31-L34) need to be filled, using:
- ArUco codedata from OpenCV: https://github.com/opencv/opencv/blob/3c3596c6efcabd0e5ee021e2048a30ec38a8d4e1/modules/objdetect/src/aruco/predefined_dictionaries.hpp#L16126-L16130


fspindle:
@s-trinh Thanks. Very nice feature.

PR #1836, opened 2025-11-25
Add headless feature for CI envs

olivier-roussel:
Introduction of `headless` variant of `opencv` package has some GUI disabled features, and should be used on CI environments only by preventing tests requiring display to be run on such environments (see failing CI here: https://github.com/lagadic/visp/pull/1834)

### DIFF
--- a/modules/gui/test/display-with-dataset/testDisplayScaled.cpp
+++ b/modules/gui/test/display-with-dataset/testDisplayScaled.cpp
@@ -329,7 +329,7 @@ int main(int argc, const char *argv[])
 #ifdef VISP_HAVE_X11
     display.push_back("X");
 #endif
-#ifdef VISP_HAVE_OPENCV && !defined(VISP_HAVE_OPENCV_HEADLESS)
+#if defined(VISP_HAVE_OPENCV) && !defined(VISP_HAVE_OPENCV_HEADLESS)
     display.push_back("OpenCV");
 #endif
 #ifdef VISP_HAVE_D3D9


olivier-roussel:
@fspindle This PR was initially fixing the use of a headless version of `opencv`, as a such package has been introduced on conda-forge. 
I've added the following:
- Added `glib` dependency that was required for `gtk2` support in ViSP (currently `gtk2` support was not enabled in ViSP)
- Cleaned deprecated dependencies: "Core Dependency Tree" (CDT) `libegl`, see [here](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html) if you want more info on what CDT is. But the important thing is that they are now removed from Conda
- `LibXML2` support was not enabled in ViSP neither as the wrong package was installed (runtime instead of development)
- Extended the support of `LibXML2` and `realsense` on Windows platforms as they now have Conda packages available on Windows

fspindle:
@olivier-roussel Thanks. I will merge this PR, but we should find a way to automatically detect whether OpenCV is headless or not in order to set `VISP_HAVE_OPENCV_HEADLESS` accordingly.

olivier-roussel:
@fspindle I agree. Unfortunately OpenCV distinguish whether it is compiled with Qt or not just at compilation stage (`WITH_QT` CMake option) and does not set anything as output to be directly used. If you diff on two OpenCV installations (one built with Qt support, the other without), you'll observe this.
The only way to detect this automatically that I see would be to extract the CMake property `IMPORTED_LINK_DEPENDENT_LIBRARIES_RELEASE` from the `opencv_highgui` target and check whether some `Qt*` targets are present or not. This is not the most elegant way but I think this is still acceptable. I'll draft a PR for this.

fspindle:
I didn't check yet, but maybe `HAVE_OPENCV_HIGHGUI` that is defined in `opencv2/opencv_modules.h` when `WITH_QT=ON` is not defined when `WITH_QT=OFF`

olivier-roussel:
No it is always defined, because all the API of highgui always exist and are identical whether it is compiled with a backend (such as Qt) or not. As I said earlier, you can check with a diff on two installs and there is absolutely no changes, which is very unfortunate.

PR #1845, opened 2025-12-02
Fast homogeneous and rotation matrix multiplications

SamFlt:
This PR introduces two methods that perform batched 3d point transformations:
- `vpHomogeneousMatrix::project`: Project N 3d points (Represented as a Nx3 vpMatrix) from a frame to another
- `vpRotationMatrix::rotateVectors`: Rotate Vectors (Represented as a Nx3)

These methods rely on explicit mat muls and SIMD operations to be faster. This PR adds tests to measure the speed up against two naive versions: the Naive `vpColVector pa = T * pb` loop and `vpMatrix::mult2Matrices` version (which multiplies a 4x4 matrix with an 4xN matrix.

Interestingly, the version without SIMD in `project` or `rotateVectors`, already provides a x2/x3 speedup, while on my machine, I can obtain x5,x6 speedups. 

Some notes:

- There is a lot of redundancy in the intrinsics inclusion at the start of the .cpp files, maybe all the relevant stuff could be put in a vpSIMDIntrinsicsUtils.h file ?
- It seems like the best way to use all the relevant CPU feature set is to use the `march=native` flag. As it is, turning on AVX with `-mavx` does not turn on FMA. This also enables all SSE relevant feature sets, which could help simplify the compileflags.


modules/core/include/visp3/core/vpSIMDUtils.h
```diff
@@ -0,0 +1,231 @@
...
+namespace vpSIMD
+{
+#if defined(VISP_HAVE_AVX2)
+using Register = __m512d;
+
+inline constexpr int numLanes = 8;
+inline const Register add(const Register a, const Register b)
+{
+  return _mm512_add_pd(a, b);
+}
```
s-trinh: Two comments.

AVX2 != 512-bits register:
- AVX / AVX2 improvements: https://fr.wikipedia.org/wiki/Advanced_Vector_Extensions#Advanced_Vector_Extensions_2
- see [AVX-512](https://fr.wikipedia.org/wiki/Advanced_Vector_Extensions#Tableau_de_compatibilit%C3%A9_des_CPU_aux_instructions_AVX-512) and all the different flavors
- you can use `cat /proc/cpuinfo` to see which AVX-512 variants your CPU has

I would definitely not expose SIMD code to the user:
- people knowing this subject will not use this code and there are already better libs doing that
- with SIMD code in `.h`, with `march=native` and running ViSP on another computer there are some chances to have SIGILL crash on old CPU
- I think what is done is runtime dispatching:
  - the `.so` can have all the more advanced instructions set
  - for example for a convolution function, a variant for 128-bits, 256-bits and 512-bits (the lib size will increase a lot)
  - and at runtime you check if the CPU has this kind of instructions set and dispatch to the more advanced one

PR #1859, opened 2025-12-12
New RBT Features

SamFlt:
This PR adds 3 new features:

- The tracked degrees of freedom of the object can be specified for each feature. For instance, Contours are not well suited to track cylindrical objects (extraneous rotation around the axis of revolution). This done in a similar fashion to what is done in the MBT.
- In `vpPointMap`, when considering points on the object (with an associated normal vector), the visible points can be filtered using a normal visibility angle criterion.
- Added an optional maximum number of features in `vpRBDenseDepthTracker`, to ensure that runtime is not too dependent on object size. We can thus use a sampling step of 1 for both small objects (using all available points then) and for larger objects (subsampling afterwards).


fspindle:
@SamFlt Thanks. Seems ok to me to merge.

SamFlt:
Hey @fspindle,

It should be good to go. I might write additional tests in another PR.

Thanks!

Issue #1861, opened 2025-12-14
Non deterministic ViSP Canny results with multithreading


s-trinh:
Hello,

Three runs with default `nbThread=-1` in `vpCannyEdgeDetection()`.
I don't know if it is expected but the result differences between three consecutive runs are visible when using default `nbThread=-1`:

<img width="512" height="340" alt="Image" src="https://github.com/user-attachments/assets/e7794aed-b207-42af-af5f-33913cf97d49" />
<img width="512" height="340" alt="Image" src="https://github.com/user-attachments/assets/ff4c941a-35a9-4b24-90a7-6b0a4bfe9c99" />
<img width="512" height="340" alt="Image" src="https://github.com/user-attachments/assets/16f3c042-8666-4877-9457-1d33e21cbe40" />

---

<details>

<summary>The code</summary>

```cpp
#include <iostream>
#include <visp3/core/vpImage.h>
#include <visp3/core/vpIoTools.h>
#include <visp3/core/vpCannyEdgeDetection.h>
#include <visp3/core/vpImageFilter.h>
#include <visp3/io/vpImageIo.h>

#ifdef ENABLE_VISP_NAMESPACE
using namespace VISP_NAMESPACE_NAME;
#endif

int main(int, const char **)
{
  std::srand(0);

  std::string input = "MEF_img_0000.png";
  vpImage<unsigned char> I_gray;
  vpImageIo::read(I_gray, input);

  int gaussianKernelSize = 3;
  float gaussianStdev = 1.0f;
  int apertureSize = 3;
  vpImageFilter::vpCannyFilteringAndGradientType filteringType = vpImageFilter::CANNY_GBLUR_SOBEL_FILTERING;
  // Canny parameters
  float lowerThresh = -1.;
  float upperThresh = -1.;
  float lowerThreshRatio = 0.6f;
  float upperThreshRatio = 0.8f;

  for (int i = 0; i < 3; i++)
  {
    vpCannyEdgeDetection cannyDetector(gaussianKernelSize, gaussianStdev, apertureSize,
                                lowerThresh, upperThresh, lowerThreshRatio, upperThreshRatio,
                                filteringType);

    vpImage<unsigned char> I_canny = cannyDetector.detect(I_gray);
    const double canny_mean = I_canny.getMeanValue();
    std::cout << "canny_mean=" << canny_mean << std::endl;

    std::ostringstream oss;
    oss << "I_canny_" << i << ".png";
[...]

PR #1862, opened 2025-12-16
[FIX] Fixed problem that occured when using several instances in parallel of vpCanny

rolalaro:
The problem came from the fact that m_edgePointsCandidates was static


fspindle:
Closes #1861

PR #1885, opened 2026-01-21
Add support for compressed NPZ file

s-trinh:
Thanks to: https://github.com/rogersce/cnpy/pull/103

---

https://github.com/lagadic/visp-images/pull/34

Reading npz from [numpy.savez_compressed](https://numpy.org/doc/stable/reference/generated/numpy.savez_compressed.html) works except somehow for string data.
Writing compressed npz file from ViSP can be read from NumPy.

Some functions should not be exposed to the user, e.g. `compressData()`, `BigEndianTest()`, `parse_npy_header()`, ..., with `VISP_EXPORT`. But removing them is not easy since I don't know how to handle this kind of thing at a library level. 

Rename to `npz_save_str()` for string parameter/data, otherwise when adding a bool parameter for compression flag:
```bash
<>/visp/tutorial/misc/npz/tutorial-npz.cpp: In function ‘int main()’:
<>/visp/tutorial/misc/npz/tutorial-npz.cpp:61:25: error: call of overloaded ‘npz_save(const std::string&, const std::string&, __gnu_cxx::__alloc_traits<std::allocator<char>, char>::value_type*, <brace-enclosed initializer list>, const char [2])’ is ambiguous
   61 |     visp::cnpy::npz_save(npz_filename, identifier, &vec_save_string[0], { vec_save_string.size() }, "w");
      |     ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from <>/visp/tutorial/misc/npz/tutorial-npz.cpp:42:
<>/visp/modules/core/include/visp3/core/vpIoTools.h:297:27: note: candidate: ‘void visp::cnpy::npz_save(const std::string&, std::string, const T*, const std::vector<long unsigned int>&, const std::string&, bool) [with T = char; std::string = std::__cxx11::basic_string<char>]’
  297 | template<typename T> void npz_save(const std::string &zipname, std::string fname, const T *data, const std::vector<size_t> &shape,
      |                           ^~~~~~~~
<>/visp/modules/core/include/visp3/core/vpIoTools.h:190:18: note: candidate: ‘void visp::cnpy::npz_save(const std::string&, const std::string&, const std::string&, const std::string&, bool)’
  190 | VISP_EXPORT void npz_save(const std::string &zipname, const std::string &fname, const std::string &data,

[...]
```

I suspect before when using `brace-enclosed initializer list` it was maybe not calling the intended function.

### DIFF
--- a/modules/core/include/visp3/core/vpIoTools.h
+++ b/modules/core/include/visp3/core/vpIoTools.h
@@ -62,9 +62,15 @@ namespace visp
 {
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
 // https://github.com/BinomialLLC/basis_universal/blob/ad9386a4a1cf2a248f7bbd45f543a7448db15267/encoder/basisu_miniz.h#L665
+// Karl Malbrain's compact CRC-32. See: "A compact CCITT crc16 and crc32 C implementation that balances processor
+// cache usage against speed": http://www.geocities.com/malbrain/
+// https://www.geocities.ws/malbrain/crc_c.html
 //
+// Other code:
 // stb crc32 implementation:
 // https://github.com/nothings/stb/blob/f1c79c02822848a9bed4315b12c8c8f3761e1296/stb_image_write.h#L1024-L1071
+//
+// https://github.com/cesanta/mongoose-os/blob/3d8bf2c1445389a710158353bf23be943ca3034d/src/common/cs_crc32.c#L20-L41
 static inline uint32_t vp_mz_crc32(uint32_t crc, const unsigned char *ptr, size_t buf_len)
 {
   static const uint32_t s_crc32[16] = { 0, 0x1db71064, 0x3b6e20c8, 0x26d930ac, 0x76dc4190, 0x6b6b51f4, 0x4db26158, 0x5005713c,


fspindle:
@s-trinh Thanks for this PR.
I just discover that lines like
```
crc = vp_mz_crc32(crc, (uint8_t *)data, nels*sizeof(T));
```
produce a warning with Xcode when building a project for iOS
```
implicit conversion loses integer precision: 'unsigned long' to 'uint_32_t' (aka 'unsigned int')
```
Since I'm not familiar with this code, could you investigate...

As a starting point I'm wondering if the function `vp_mz_crc32()` should not return an `uint_32_t` and use internally also `uint_32_t` instead of an `unsigned int`

```
static inline uint_32_t vp_mz_crc32(unsigned long crc, const unsigned char *ptr, size_t buf_len)
{
  static const unsigned int s_crc32[16] = { 0, 0x1db71064, 0x3b6e20c8, 0x26d930ac, 0x76dc4190, 0x6b6b51f4, 0x4db26158, 0x5005713c,
    0xedb88320, 0xf00f9344, 0xd6d6a3e8, 0xcb61b38c, 0x9b64c2b0, 0x86d3d2d4, 0xa00ae278, 0xbdbdf21c };
  uint_32_t crcu32 = static_cast<uint_32_t>(crc);
  if (!ptr) return 0;
  crcu32 = ~crcu32;
  while (buf_len--) {
    unsigned char b = *ptr++;
    crcu32 = (crcu32 >> 4) ^ s_crc32[(crcu32 & 0xF) ^ (b & 0xF)];
    crcu32 = (crcu32 >> 4) ^ s_crc32[(crcu32 & 0xF) ^ (b >> 4)];
  }
  return ~crcu32;
}
```

s-trinh:
https://github.com/lagadic/visp/blob/838e63abaa44272abade73ad6affa55ff65c8242/modules/core/include/visp3/core/vpIoTools.h#L64-L78

---

`unsigned long` type can be 64-bits on some platform. I have pushed a commit to use `uint32_t` instead.

https://en.cppreference.com/w/cpp/language/types.html

image

---

stb implementation of crc32:
https://github.com/nothings/stb/blob/f1c79c02822848a9bed4315b12c8c8f3761e1296/stb_image_write.h#L1024-L1071

fspindle:
@s-trinh Thanks

PR #1921, opened 2026-03-02
Force enable visp namespace to ON

olivier-roussel:
According to discussion in our last ViSP dev meeting, we should enable by default the visp namespace.
Some documentation will have to be updated.


fspindle:
@SamFlt Thanks for your time and help. It builds and works on my macOS. Waiting for ci...

PR #1953, opened 2026-05-18
Fix shadow warnings

fspindle:


example/device/framegrabber/grab1394Two.cpp
```diff
@@ -482,7 +482,7 @@ int main(int argc, const char **argv)
               for (it_lcoding = lcoding.begin(); it_lcoding != lcoding.end(); ++it_lcoding) {
                 vp1394TwoGrabber::vp1394TwoColorCodingType supcoding;
                 supcoding = *it_lcoding;
-                std::stringstream ss;
+                std::stringstream sss;
```
s-trinh, line 485: Very quick glance over the changes.
`sss` seems to be never used after?

---

Current enabled compiler flags if available:

https://github.com/lagadic/visp/blob/db55dcbcda81d7231101a9e44a43771df1af9a72/cmake/AddExtraCompilationFlags.cmake#L81-L100

Maybe two ideas:
- for the available GHAs
  - pass some additional compiler flags, probably using `-DCMAKE_CXX_FLAGS` (not sure if `-DCMAKE_C_FLAGS` should also be used or not)
  - `-Wall -Wextra -Wshadow -Wnon-virtual-dtor -pedantic` --> https://github.com/cpp-best-practices/cppbestpractices/blob/master/02-Use_the_Tools_Available.md#gcc--clang
  - I don't know if `-Wpedantic` should also be added or used instead ; its purpose
  - the doc: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-pedantic-1
  - save the build log in a txt file (something like `make -j8 2>&1 | tee build.log >(grep -i warning > warnings.txt)`) and cat all into a separate section in the github action to be able to read only the warnings and not the full build log
  - this should not significantly impact the build time
  - some other links from google search:
    - https://bytes.usc.edu/cs104/wiki/gcc/
    - https://dvergeylen.github.io/gcc-most-common-used-flags (here there is a warning "Never put -Wpedantic in production code, as the outputted warnings can vary from standard to standard. [...]")
    - https://stackoverflow.com/questions/2855121/what-is-the-purpose-of-using-pedantic-in-the-gcc-g-compiler
- add another GHA:
  - latest Ubuntu
  - must run in less than maybe 5 min
  - only the essential dependencies (OpenCV) installed from APT
  - use `-Werror` to have a build failure since the build logs are read only when there is an issue
  - probably `-pedantic` should not be used here since only "important" warnings are needed

PR #1972, opened 2026-06-24
Add build fails if warnings in GHA

s-trinh:


s-trinh:
@fspindle 

I have added a commit with `set -o pipefail` to make it fails directly in the build step if there is a compilation error.
To avoid this issue:

image

PR #1981, opened 2026-07-07
TensorRT tutorial

rolalaro:
Updated the code to be able to compile and run the turorial at least with the following configuration:
- Cuda 12.4
- cuDNN 9.23
- TensorRT 11.0
- OpenCV 4.13

Updated the installation instructions to rely on Conda, which is a more robust way to handle Cuda and cuDNN


fspindle:
@rolalaro Thanks

PR #1998, opened 2026-08-12
Attempt to fix catchParseArgv (Exit code 0xc0000374) on Windows

rolalaro:
Might be a problem of order of operations on Windows


rolalaro:
The failed CIs are not related anymore to Windows, which means that this PR's goal has been reached. Consequently, this PR will be merged and a new one will be open to address the problem of the Python test that fails

PR #2023, opened 2026-09-16
Tutorial camera realsense

sjourdro:
Here is the tutorial on how to use a RealSense camera.
I am still adding a part on how to get a grayscale image and a depth representation.


sjourdro:
@rolalaro @fspindle 

The tutorial is ready for review.

rolalaro:
Great job ! =)
I would also have added either a reference to an already existing examples where both color and depth images are retrieved, or a new short example with such a configuration. What do you think @sjourdro @fspindle ?

fspindle:
@rolalaro This is exactly what I asked this morning ;-)

sjourdro:
The dual view example has been added.

@fspindle @rolalaro

